PR #121, opened 2015-01-05
safer access to list element's data

monsta:
g_list_nth_data just returns NULL if either list or element is NULL


infirit:
Merged

PR #123, opened 2015-01-10
comics backend: add support for bsdtar

City-busz:
bsdtar is able to extract all supported comics archive format. (Even though 7z support is slow.) The command line parameters are same as tar.


infirit:
Merged in d8d4fcd1182e4295cb4307333a4af4ea3e3c75db.

Thx

PR #134, opened 2015-04-20
Made the dependancy on mate-desktop optional

brakarov:
Hi,

Since I use XFCE I don't like the fact that mate-desktop is installed on my system just to support Atril. Evince looks out of place on my desktop since CSD so I opted for Atril.

Because Evince didn't require gnome-desktop I guessed that the work involved wouldn't be that much. So rather than complain on the fora I decided to do something about that.

I'd guess this would make Atril more interesting for XFCE users.

Regards,
Geert

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -199,7 +199,7 @@ case "$with_platform" in
         ;;
 esac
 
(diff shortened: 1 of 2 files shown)


rilian-la-te:
+1 to this. I also want to use atril in my GTK3 LXDE, but does not want to build half of mate for it.

infirit:
Merged with small correction in spelling of MATE.

Issue #136, opened 2015-05-11
Segfault after pressing ESC with no document opened


tu-ghgoso:
The buggy code is inside ``ev_window_cmd_escape'':
   if (!window->priv->document->iswebdocument && window->priv->view)
      ev_view_autoscroll_stop (EV_VIEW (window->priv->view));

``window->priv->document'' might actually be NULL.

PR #142, opened 2015-07-10
fix crash when pressing Esc with no document opened

monsta:
fixes https://github.com/mate-desktop/atril/issues/136


infirit:
merged

PR #151, opened 2015-08-24
drop support for win32/osx/hildon, make smclient mandatory

monsta:


flexiondotorg:
Did you really mean to drop the dbus backend?

monsta:
Well we can leave it but as I read somewhere it has less features than XSMP one.
It only affects smclient though, so if I'm mistaken here, I can always add it back.

monsta:
@flexiondotorg: actually, [eggsmclient-dbus.c](https://git.gnome.org/browse/libegg/tree/libegg/smclient/eggsmclient-dbus.c) was never here, and was never mentioned in configure.ac. So we're good to go here.

infirit:
merged

PR #155, opened 2015-08-27
configure: properly show "epub backend: no" in summary if it's off

monsta:
also fixed some indentation issues


infirit:
pushed slightly different fix

PR #168, opened 2015-10-31
1.12 the ability to work as previewer for vim-latex users

gnugr:
Please consider the ability for atril to work on MATE as previewer for LaTeX users on vim editor.
vim-latex-live-preview is a plugin for vim as it shown there:
https://github.com/xuhdev/vim-latex-live-preview
With this ability atril could be one of the best previewers.


raveit65:
Where is your commit for this PR?

PR #207, opened 2016-06-13
shell: Respect user's toolbar-style setting

zeldin:
Atril should respect if the user wants another toolbar-style than "both-horiz" by checking the "toolbar-style" key of the "org.mate.interface" schema in GSettings.

### DIFF
--- a/shell/ev-window.c
+++ b/shell/ev-window.c
@@ -1487,7 +1487,8 @@ ev_window_setup_toolbar_flags (EvWindow *ev_window)
 {
 	EggTbModelFlags flags = egg_toolbars_model_get_flags(ev_window->priv->toolbars_model, 0);
 	char *style;
-	if ((style = g_settings_get_string (ev_window->priv->interface_settings, MATE_INTERFACE_TB_STYLE))) {
+	if (ev_window->priv->interface_settings &&
+	    (style = g_settings_get_string (ev_window->priv->interface_settings, MATE_INTERFACE_TB_STYLE))) {
 		flags &= ~EGG_TB_MODEL_STYLES_MASK;
 		if (!strcmp (style, "both")) {
 			flags |= EGG_TB_MODEL_BOTH;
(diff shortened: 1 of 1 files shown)


monsta:
Nice - but would be better to check the existence of the schema before using it. There are people who use Atril in other desktop environments.

Also this setting is completely ignored in GTK+3 build - they made it no-op at some point... IIRC it was in GTK+ 3.10.

zeldin:
Right you are.  I run it in XFCE myself, but I still have all the MATE schemas installed.  :smile:
I pushed a fix to this PR.

> Also this setting is completely ignored in GTK+3 build

Hm?  The patch works fine for me (as in, all the styles do what they are supposed to when I set the setting) with GTK+3.20.6 on Debian stretch...

monsta:
I recall eom ignores this setting in GTK+3 build, so I thought Atril does the same... I'll check it once more.

monsta:
Hmm, interesting... GTK+2 build of Atril reacts to this setting even without this PR. The only issue is that "icon" value isn't processed properly - Atril uses "both" value instead. GTK+2 build of eom has the same behavior. I'll check GTK+3 build later.

raveit65:
@monsta 
What is the state of merging this PR?
Or do we need this with gtk3 build any more?

monsta:
Ok, it works with current git master. I guess it's because we have a custom "libegg" toolbar, not some standard GTK+ one. I remember the toolbars settings were ignored in the standard one, therefore we removed them from the last tab of mate-appearance-properties...

monsta:
> GTK+2 build of Atril reacts to this setting even without this PR. The only issue is that "icon" value isn't processed properly - Atril uses "both" value instead. 

Here in GTK+3 code we get the same behavior with this PR. In other words, it seems ok since it restores the old behavior. As for some values not working, I guess it's some separate issue in the toolbars.

raveit65:
hmm, it seems to be a bit inconsistent.
- both --> text is under icons on most buttons
- icon --> i see only icons, but for forward/previous buttons i see icons and text.
- both-horz --> same as with icons

monsta:
Yes, this inconsistency is what I was talking about. You can check GTK+2 build and see it there as well. That's why I think it's some separate issue in the toolbars.

raveit65:
As, this doesn't work for normal Gtk Toolbars i am in doubt that i will ever use that ;-)
I don't like to use all the old deprecated gtk+2 stuff for myself........
So feel free to merge it if you think it is OK.

monsta:
Ok, rebased and merged.

PR #210, opened 2016-06-29
deprecation fixes and backports from evince round 2

raveit65:
@monsta @flexiondotorg 
Please test and review.


monsta:
Works alright in GTK+2. I'm not sure why upstream added Ctrl-Up/Down bindings for prev/next search result, because we already have Ctrl-G/Shift-Ctrl-G and F3... but since they work fine here, I don't mind.

Issue #211, opened 2016-06-30
(pdf) selected text -> right button -> black text


sc0w:
debian atril 1.14.1 gtk3

I selected text in a pdf, then right button on the selected text, and shows black text

The epub format doesn't have this bug

raveit65:
gtk+-3.20 --> maybe a theme issue --> but without more information --> i'm clueless ;)

sc0w:
I use the theme TraditionalOk

gtk 3.20.6

I see this only in atril

sc0w:
exactly same than https://github.com/linuxmint/xreader/issues/19 if the window loses focus

raveit65:
A style class .view is missing, here i'm working on it.
For the moment i fixed that in themes for gtk+-3.20
https://github.com/mate-desktop/mate-themes/commit/9b6cd516e0ff7a62eafb6edcf6641df2cb9ddbc2
@lukefromdc
I would do the same for BlackMate.
Here the 'active' bg color isn't defined.
But i don't know which colors you prefer if window is unfocused.

lukefromdc:
Use the same color text as on the normal focussed window. I don't 
generally change text colors on unfocussed windows simply because
multi-window workflows are normal for me and I prefer contents not
to change too much between focussed and not.

sc0w:
@lukefromdc I agree with you

epub format uses the same color text as the normal focussed window

@raveit65 thanks, your commit works, but the color is too light

![before](https://cloud.githubusercontent.com/assets/7734191/16536165/c04d0798-3fee-11e6-83aa-04628585b08f.png)

![later](https://cloud.githubusercontent.com/assets/7734191/16536168/c9fab90c-3fee-11e6-890b-c076b4b5da92.png)

raveit65:
which color?
selected bg?
selected fg?
active bg?
active fg?

raveit65:
btw. ebub is a zip container, here you can change nothing

raveit65:
@sc0w I'm complete passionless with that theme as for my taste it is boring and something from 1884.
...only stupid work for me.
Do you like to maintain it?

sc0w:
this grey color

![grey](https://cloud.githubusercontent.com/assets/7734191/16536942/51e84d66-3ff9-11e6-8e23-8663603c24cd.png)

Is only a suggestion

Thanks for solve this issue quikly :)

raveit65:
Well, i can make the active bg color darker.
But what about the contrast than?

[...]

PR #212, opened 2016-07-03
improvements and warnings

raveit65:
@monsta @flexiondotorg @lukefromdc
Please test and review.

Btw.
Adding the style classes is the real fix for https://github.com/mate-desktop/atril/issues/211

### DIFF
--- a/cut-n-paste/gimpcellrenderertoggle/gimpcellrenderertoggle.c
+++ b/cut-n-paste/gimpcellrenderertoggle/gimpcellrenderertoggle.c
@@ -61,7 +61,11 @@ static void gimp_cell_renderer_toggle_set_property (GObject         *object,
                                                     GParamSpec      *pspec);
 static void gimp_cell_renderer_toggle_get_size     (GtkCellRenderer *cell,
                                                     GtkWidget       *widget,
+#if GTK_CHECK_VERSION (3, 0, 0)
+                                                    const GdkRectangle *rectangle,
+#else
                                                     GdkRectangle    *rectangle,
+#endif
                                                     gint            *x_offset,
                                                     gint            *y_offset,
                                                     gint            *width,
@@ -86,8 +90,13 @@ static gboolean gimp_cell_renderer_toggle_activate (GtkCellRenderer *cell,
                                                     GdkEvent        *event,
                                                     GtkWidget       *widget,
                                                     const gchar     *path,
+#if GTK_CHECK_VERSION (3, 0, 0)
+                                                    const GdkRectangle *background_area,
+                                                    const GdkRectangle *cell_area,
+#else
                                                     GdkRectangle    *background_area,
                                                     GdkRectangle    *cell_area,
+#endif
                                                     GtkCellRendererState  flags);
 static void gimp_cell_renderer_toggle_create_pixbuf (GimpCellRendererToggle *toggle,
                                                      GtkWidget              *widget);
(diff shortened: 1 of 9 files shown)


lukefromdc:
Very nice work, I just built this and it was good enough to be worth packing
up into a .deb package. Always I'd seen that selected text issue but just
thought of it as one of the quirks of PDF readers.

libview/ev-view.c
```diff
@@ -3516,6 +3521,10 @@ ev_view_scroll_event (GtkWidget *widget, GdkEventScroll *event)
 		        case GDK_SCROLL_LEFT:
 				ev_view_previous_page (view);
 				break;
+#if GTK_CHECK_VERSION (3, 0, 0)
+                        case GDK_SCROLL_SMOOTH:
+                                g_assert_not_reached ();
```
monsta: This is going to make Atril crash on receiving ```GDK_SCROLL_SMOOTH``` event (which can happen in GTK+3 build, I already encountered such crash in some other app). Need to backport proper handling from upstream...

cut-n-paste/gimpcellrenderertoggle/gimpcellrenderertoggle.c
```diff
@@ -61,7 +61,11 @@ static void gimp_cell_renderer_toggle_set_property (GObject         *object,
                                                     GParamSpec      *pspec);
 static void gimp_cell_renderer_toggle_get_size     (GtkCellRenderer *cell,
                                                     GtkWidget       *widget,
+#if GTK_CHECK_VERSION (3, 0, 0)
+                                                    const GdkRectangle *rectangle,
+#else
                                                     GdkRectangle    *rectangle,
```
monsta: It should work with any GTK+ version.

libview/ev-view.c
```diff
@@ -312,7 +326,8 @@ static void       ev_view_primary_clear_cb                   (GtkClipboard
 static void       ev_view_update_primary_selection           (EvView             *ev_view);
 
 #if GTK_CHECK_VERSION (3, 0, 0)
-G_DEFINE_TYPE_WITH_CODE (EvView, ev_view, GTK_TYPE_FIXED, G_IMPLEMENT_INTERFACE (GTK_TYPE_SCROLLABLE, NULL))
+G_DEFINE_TYPE_WITH_CODE (EvView, ev_view, GTK_TYPE_CONTAINER,
+                         G_IMPLEMENT_INTERFACE (GTK_TYPE_SCROLLABLE, NULL))
```
monsta: I wonder if we can make EvView derive from GtkContainer in GTK+2 build as well. This would make all this code a bit simpler and easier to read.

monsta:
Overall, in GTK+2 build it seems to work fine here.

cut-n-paste/gimpcellrenderertoggle/gimpcellrenderertoggle.c
monsta: (I mean all the changes in this commit)

raveit65: gtk2 build don't give a warning here, so i make it gtk3 only.
Personal i don't want to touch gtk2 code if it isn't necessary or needed, on the last miles for migration to  gtk3.
Well, let us talk at irc about.

libview/ev-view.c
raveit65: in line 3574 gtk_style_context_set_background is called, don't know if the rest of the new code is GTK releated

libview/ev-view.c
raveit65: Well, i can look at upstream if there are some followers.
But, if i understand upstream commit correct is that a dummy function to fix the build warning.

monsta: My bad, didn't notice the ```return``` statement was added several lines before it. This line won't be reached indeed.

cut-n-paste/gimpcellrenderertoggle/gimpcellrenderertoggle.c
monsta: Great, now I get warnings with GTK+2... stupid undocumented API changes. You're right, let's leave it as is.

PR #215, opened 2016-07-22
backports

raveit65:
@monsta @flexiondotorg
please test, now we are at state of evince-3.4.0


monsta:
I'm not sure why, but once I select either Dual (Even pages left) or Dual (Odd pages left) item in the View menu, I can't go back to single page view mode... I can only switch between two dual-page modes. Did I miss some menu item for that? :confused:

monsta:
Ah... got it... just need to click the current mode item once more, the checkbox goes away, and single page mode is back :smile:

monsta:
Well, seems to run fine with GTK+2 here.

PR #224, opened 2016-10-16
Fix case sensitive searching for PDF

cormac-w:
Poppler has a facility for case sensitive search, but it is not being used at the moment by atril. This patch calls the poppler search function with the right options.


monsta:
Thanks, merged :smile:

PR #226, opened 2016-10-21
djvu: support image/vnd.djvu+multipage MIME type

monsta:
more info at:
https://git.gnome.org/browse/evince/commit/?id=7850b986cf3225369170cdec545844b025bfd12e

This fixes https://bugs.debian.org/840324. Link to .djvu file for testing is also in that report.

@raveit65 @lukefromdc @sc0w guys, if you can, please test it on various (old and new) distro releases. The same .djvu file should now open fine with any version of shared-mime-info package (see the upstream commit for explanation).


lukefromdc:
I can now confirm this works-and that it did not work without this PR. I support merging it.

PR #237, opened 2017-01-08
force X11

raveit65:
atril runs with this warning in gnome-wayland with this PR
```
[rave@f26 ~]$ atril
PlatformDisplayWayland initialization: failed to connect to the Wayland display: 
```
...whatever this means.
But it runs.

### DIFF
--- a/shell/main.c
+++ b/shell/main.c
@@ -234,6 +234,8 @@ main (int argc, char *argv[])
 	textdomain (GETTEXT_PACKAGE);
 #endif
 
+	gdk_set_allowed_backends ("x11");
+
 	context = g_option_context_new (N_("MATE Document Viewer"));
 	g_option_context_set_translation_domain(context, GETTEXT_PACKAGE);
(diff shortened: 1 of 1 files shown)


monsta:
Well, it runs as before on X, so it's ok for me. Guess we need to ask @lukefromdc to test it with Wayland.

lukefromdc:
A test build and install worked in Weston, though right now (current Weston) Atril and EOM have to be launched from terminal and the desired file opened from the open dialog. Caja works but attempting to open something from Caja in the usual way fails. Might be specific to Weston and not to Wayland. a GNOME wayland session is exceedingly difficult for me to get to as it requires setting GDM as the display manager first. Lightdm cannot handle Wayland session, I am launching Weston from the console directly

raveit65:
merged
https://github.com/mate-desktop/atril/commit/4b264c661a53e847b958cf6c8913c88a8cc707d5

PR #240, opened 2017-02-20
First page should not reloaded

nanxiongchao:
For the displayed page, do not reload every time, should be displayed immediately
Tested successfully on Loongson 3A2000
with fedora25(mips64el distribution).

### DIFF
--- a/libview/ev-pixbuf-cache.c
+++ b/libview/ev-pixbuf-cache.c
@@ -90,7 +90,7 @@ static gboolean      new_selection_surface_needed(EvPixbufCache      *pixbuf_cac
 #define PAGE_CACHE_LEN(pixbuf_cache) \
 	(pixbuf_cache->start_page>=0?((pixbuf_cache->end_page - pixbuf_cache->start_page) + 1):0)
 
-#define MAX_PRELOADED_PAGES 3
+#define MAX_PRELOADED_PAGES 20
 
 G_DEFINE_TYPE (EvPixbufCache, ev_pixbuf_cache, G_TYPE_OBJECT)
 
--- a/libview/ev-view.c
+++ b/libview/ev-view.c
@@ -732,12 +732,13 @@ view_update_range_and_current_page (EvView *view)
 		}
 	}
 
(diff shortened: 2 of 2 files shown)


jinguojie-loongson:
I guess this patch is aiming to enlarge the size of the page cache.
Suppose we open a large pdf file. Besides the first page is rendered and displayed, several following pages are also rendered in the background. Then we scroll down to the second page, and it's displayed without extra rendering time.
Surely this will improve user experience.
Looks fine to me.

raveit65:
What is the advantage for not auto-reloading the first page?
Ie. another application change the content of the document, maybe the first page.

jinguojie-loongson:
@raveit65 
We observed that some complex PDF files need several seconds to render a page.
As an example, a printed version of a magazine usually contains large page sizes, lots of vectorized shapes, hierarchical layers, transparent objects, and many other advanced effects. All of them take high volume of computing time.
To disable the auto-reloading and enlarge the cache, we can reduce waiting time and browse the pages more smoothly.
Since Atril is considered as a "document viewer", how about treat the PDF file as “readonly” in most cases?

raveit65:
Nice, i just tested it with that large pdf https://www.nasa.gov/connect/ebooks/hubble_25_detail.html
W/o the PR atril reload the firstpage if i scroll back from page 4.
With the PR i can scroll back from page 21 without reloading page one.
Thank you, but we need another reviewer to approve the PR.

raveit65:
merged https://github.com/mate-desktop/atril/commit/00fad2c0d0fcdade570a08f046e85a4558f1fc0c
Thank you.

raveit65:
@nanxiongchao @jinguojie-loongson 
Sadly this commit have a heavy side effect  and increases meomory consumption, especially with HIDPI displays. In my tests atril can use > 4GB memory with default page-cache=50MB.
When increasing page-cache mem consuption can be much higer and a system with 16 GB mem can run out of memory easily.
See this report for more info https://github.com/mate-desktop/atril/issues/404
We need to find a solution with your commit, reverting it fixes the high memory load.

PR #242, opened 2017-03-04
Previously opened page incorrectly restored if it was the last one

nanxiongchao:
If you stay on the last page when you close it, reopen it will
not be positioned until you visited it (reset to the first).
In the changed code:
Judgment only works on the last page.
This function explicitly introduces this problem, so it can be
removed directly.
I have been tested successfully.


raveit65:
merged https://github.com/mate-desktop/atril/commit/b2b6dc9afb1c5057415e8eb08a230d18ee955e99
Thank you

PR #244, opened 2017-03-05
replace several deprecations

raveit65:
please test


monsta:
Findbar now expands vertically, taking up all the height of the bottom bar. Is that ok?

Before:
atril-1

After:
atril-2

raveit65:
I didn't set a vertical expand here, but i noticed that too.
For me it looks OK.
Any ways, i will try to add a vertical margin here.

raveit65:
@monsta 
Now it looks like before.
I will squash that if you agree ;)

PR #245, opened 2017-03-07
first patch set of backporting accessiblity improvements from upstream

raveit65:
There are a lot more accessiblity commits to backport from upstream, but i will stop at this point to make the change not to huge.
Next step will be include carret navigation with atk commit, but this will come later.

### DIFF
--- a/libdocument/ev-mapping-list.c
+++ b/libdocument/ev-mapping-list.c
@@ -60,6 +60,15 @@ ev_mapping_list_find_custom (EvMappingList *mapping_list,
 	return NULL;
 }
 
+EvMapping *
+ev_mapping_list_nth (EvMappingList *mapping_list,
+                     guint          n)
+{
+        g_return_val_if_fail (mapping_list != NULL, NULL);
+
+        return (EvMapping *)g_list_nth_data (mapping_list->list, n);
+}
+
 gpointer
 ev_mapping_list_get_data (EvMappingList *mapping_list,
 			  gdouble        x,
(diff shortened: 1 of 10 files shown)


raveit65:
@ikeydoherty
please test

cwendling:
@raveit65 I did a quick test, and it seems fairly good in Accerciser, that is it seems to provide the appropriate interfaces (text, links, etc.).  However, I'm not sure how I'm supposed to actually navigate the document in a useful way for an AT: Evince has a keyboard mode that can be triggered with <kbd>F7</kbd>, and that then allows both navigating the document and the screen reader to know what it should read, but Atril doesn't seem to.  So as is it seems a better, but not yet usable.

Also, I think there is a memory bug in there somewhere: I didn't take time to debug it, but the first time I browsed to the document in Accerciser Atril crashed, and when run under Valgrind and browsing to it in Accerciser it shows a memory error (in ATK, but I guess it comes originally from somewhere else).

Finally, I also noticed that the Index sidebar doesn't announce the actual element but only the page number.  Evince is fixed in that regard.

raveit65:
>  I did a quick test, and it seems fairly good in Accerciser, that is it seems to provide the appropriate interfaces (text, links, etc.). However, I'm not sure how I'm supposed to actually navigate the document in a useful way for an AT: Evince has a keyboard mode that can be triggered with F7, and that then allows both navigating the document and the screen reader to know what it should read, but Atril doesn't seem to. So as is it seems a better, but not yet usable.

Yes, there are tons of commits in evince which needs to back ported, my plan is to browse through evince history and start pick up the oldest one, and  than i will back port newer ones.
My hope is that  all is fixed after the process.
For that i need your help from time to time to verify that i do not introduce mistakes.

> Also, I think there is a memory bug in there somewhere: I didn't take time to debug it, but the first time I browsed to the document in Accerciser Atril crashed, and when run under Valgrind and browsing to it in Accerciser it shows a memory error (in ATK, but I guess it comes originally from somewhere else).

This needs to be debug by someone other, i am not able to use valigrind. I can only back port stuff from upstream, real coding is far from my knowledge.

> Finally, I also noticed that the Index sidebar doesn't announce the actual element but only the page number. Evince is fixed in that regard.

Thanks for detailed testing.

PR #250, opened 2017-04-03
deprecations

raveit65:
Please test.

### DIFF
--- a/cut-n-paste/toolbar-editor/egg-editable-toolbar.c
+++ b/cut-n-paste/toolbar-editor/egg-editable-toolbar.c
@@ -298,7 +298,13 @@ move_item_cb (GtkAction          *action,
   gdk_event_get_coords (realevent, &event->motion.x, &event->motion.y);
   gdk_event_get_root_coords (realevent, &event->motion.x_root, &event->motion.y_root);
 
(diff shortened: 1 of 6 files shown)


raveit65:
You can create a pdf with libreoffice with a password.
The message area you see if you don't file out a password if reguested during opening a document..

monsta:
I'm not sure why toolbar editor has a horizontal separator inside its window, but that separator is vertically expanded on window resize, which doesn't look very good...

Default window size:
atril-1

Resized window:
atril-2

monsta:
Other stuff seems to be ok, including the password dialog.

raveit65:
For some reason this does not happen with submarine themes, which i use mostly.
But, confirmed.....

raveit65:
@monsta 
Separator is fixed.

monsta:
Who could've thought you can expand separator in the unusual direction :slightly_smiling_face:

raveit65:
It was simply a copy&paste typo, it was never expand vertical with GtkAlignment.
```
-          alignment = gtk_alignment_new (0.5, 0.5, 1.0, 0.0);
```
But yes, the separator is a bit weird. In css i can set a background, border, and vertical/horizontal size.
For this reason i didn't saw the issue with submarine themes as here was only the border defined. :slightly_smiling_face:

PR #261, opened 2017-07-19
comics: make the files containing "--checkpoint-action=" unsupported

### LINKED ISSUE #257: Command injection vulnerability in CBT handler

sc0w:
Fixes #257 

.cbt file for testing:

[sample.zip](https://github.com/mate-desktop/atril/files/1158641/sample.zip)

### DIFF
--- a/backend/comics/comics-document.c
+++ b/backend/comics/comics-document.c
@@ -983,6 +983,12 @@ extract_argv (EvDocument *document, gint page)
 	char *command_line, *quoted_archive, *quoted_filename;
 	GError *err = NULL;
 
+	if (g_strrstr (comics_document->page_names->pdata[page], "--checkpoint-action="))
+	{
+		g_warning ("File unsupported\n");
+		gtk_main_quit ();
+	}
+
         if (page >= comics_document->page_names->len)
                 return NULL;
 


andreygursky:
> A graphical warning would be nice, but is not necessary to protect people's computers.

If something doesn't work (program closes), open it in the terminal and search for warnings. Fair trade-off.

> Also most people would assume that a particular file that causes the program used to open it to crash is defective...

Not agreed. Crash means the program is defective. If a program has problems, it could close without graphical infos but must not crash.

lukefromdc:
A sudden closure of this type looks the same as a crash to an end user.

sc0w:
the PR works fine with

`return NULL;`

instead

`gtk_main_quit ();`

and the gui is working, @lukefromdc @andreygursky you can test that

but, I noticed some warnings in the terminal, and for security reasons I think it is safely just quickly to close atril

@monsta @raveit65 

what do you think?

lukefromdc:
I can confirm that using 
`return NULL;`
keeps the GUI open and the test file does not appear in /tmp. More user friendly so long as no varient of the attack can run prior to that point. If this plugs the hole its what we should do.

lukefromdc:
The warnings do not seem to relate to the attack, just to attempting to open an empty file. Whether we use this method should this depend on how confident we are this always blocks the attack, as an attack in progress is an emergency. Pulling the plug is fine and few well see what looks like a crash but is not if we go that way, so I am fine with this either way

raveit65:
> the PR works fine with
`return NULL;`
instead
`gtk_main_quit ();`
and the gui is working,
but, I noticed some warnings in the terminal, and for security reasons I think it is safely just quickly to close atril
what do you think?

I think the user should know that something was happened in the background, but please without trigger a bug-reporting tool like we have in unbuntu or fedora.
So, if a second run from terminal gives us an info, than i think it's better to let atril close.
Assuming that a Linux user always starts the application in a terminal at an unexpected end of the program. ;-)
But i am undecided, up to you guys......

Btw. Can i close the other PR ?

lukefromdc:
I think the other PR can be closed, and that we do need to absolutely avoid triggering automatic bug reporting tools. Security is a feature, not a bug...

sc0w:
yes, the other PR can be closed

I like to see we can solve issues without deleting features

I think to close immediately atril is the best choice, and, in the future we can identify the killer files with this behavior

raveit65:
So, is this PR ready to go?

lukefromdc:
Seems to me this is ready

raveit65:
merged https://github.com/mate-desktop/atril/commit/f4291fd62f7dfe6460d2406a979ccfac0c68dd59
Thank you

lukefromdc:
Just posted a security update version of my Atril debian package to
https://archive.org/details/DebianPackagesForMate-desktopWityGtk3AndCustomPanelTheme
I would recommend cherrypicking this back to 1.18 since that includes the unarchiver support, so distros can pick it up. The exploit is publicly known, any potential attacker watching us or watching GNOME now knows how to do this.

PR #273, opened 2017-08-12
Back ports basicly from evince for the annotation window and sidebar

raveit65:
Please test.

### DIFF
--- a/atril-document.h
+++ b/atril-document.h
@@ -56,6 +56,7 @@
 #include <libdocument/ev-selection.h>
 #include <libdocument/ev-transition-effect.h>
 #include <libdocument/ev-version.h>
+#include <libdocument/ev-macros.h>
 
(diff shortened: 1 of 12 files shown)


raveit65:
> There is a jump in the main toolbar on moving from page 9 to page 10 but that is old and has been in master for a while, so does not relate to this PR.

That's the label from page number GtkEntry, which moves the icons right from it, if it displays (9from50) --> (10from50).
Isn't that a issue?
Or do you mean something different?
In any case i didn't touched the toolbar with this PR.

lukefromdc:
It does nor affect this PR in any way, this seems ready to go.  The page number you mentioned is exactly the issue I found, since it has not been reported yet I opened it as 
https://github.com/mate-desktop/atril/issues/275

raveit65:
Only a rebase against master, nothing is changed.

Issue #255, opened 2017-06-19
Thumbnails in sidebar disappear after fast scrolling


Photon89:
#### Expected behaviour

Thumbnails should be always visible

#### Actual behaviour

Thumbnails sometimes disappear after fast scrolling

#### Steps to reproduce the behaviour

Open a PDF with many pages (I tested with a PDF of 140 pages). Scroll down within the sidebar. The thumbnails might take a  bit to get shown but they will get shown eventually. However, if you scroll fast, the thumbnails will disappear. They also won't appear if you stop scrolling and wait. To make them visible you have to scroll slowly till you reach some thumbnails which are actually shown. You can then slowly scroll back to see the previously invisible thumbnails. But they will disappear again if you scroll quickly. 

edit: Just noticed: You can make the hidden thumbnails visible by touching (hovering) the separation line between the sidebar and the main PDF with the cursor as well.

#### MATE general version

None, I use Atril in Xfce

#### Package version

1.18.0

#### Linux Distribution

Arch Linux

Similar bug in evince: https://bugzilla.gnome.org/show_bug.cgi?id=701537

oliveiraleo:
@Photon89 I have the same issue. I'm on Arch Linux too, but using MATE 1.18.0 and Atril (1.18.0) has same thing.
My tip is to not maximize the PDF window, it works for me in thumbnail disappearing.

TomaszGasior:
I can confirm this bug on Atril. Evince is not affected. Please see my movie with comparison Evince vs Atril.
[vokoscreen-2017-07-23_14-35-52.mp4.zip](https://github.com/mate-desktop/atril/files/1168070/vokoscreen-2017-07-23_14-35-52.mp4.zip)

Manjaro Linux
atril 1.18.0-1
evince 3.24.0+12+g717df38f-1

Photon89:
It's interesting that three people are hit by this issue and all of us use Arch/Manjaro. Maybe a packaging problem in the end? Here is Atril's Arch package: https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/atril

raveit65:
Same in fedora 26.

TomaszGasior:
I found that this bug exists in original Evince in version 3.24.0 but it is fixed in 3.24.1. Can someone find a commit that fixes this bug and apply to Atril?

Photon89:
[...]

PR #274, opened 2017-08-13
sidebar thumbnails

raveit65:
Fixes https://github.com/mate-desktop/atril/issues/255
Please test.


lukefromdc:
This is so valuble a fix I decided to pack this up as a Debian package immediately, as I don't want to go back to those disappearing thumbnail issues that made sense on a machine with 128MB of RAM but don't make sense today.

raveit65:
only rebased against master.

PR #276, opened 2017-08-15
more backports

raveit65:
Please test.
Some commits needs to test with a touchpad.


raveit65:
I don#t like to use a touchpad verymuch so i have no idea how 
pinch/zoom gesture
should work.
Any ways, i see no regressions with my notebook and i want trust linuxmint/gnome devs.

PR #278, opened 2017-08-24
improve annotations

raveit65:
Basicly this PR does make 2 things:
- set transparancy for the annotation window too
- annotations can be removed now

A lot of commits before were needed to make this working and to let all clean apply.

### DIFF
--- a/shell/ev-window.c
+++ b/shell/ev-window.c
@@ -5361,7 +5361,8 @@ view_menu_annot_popup (EvWindow     *ev_window,
 {
 	GtkAction *action;
 	gboolean   show_annot = FALSE;
-	gboolean can_remove_annots;
+	gboolean can_remove_annots = FALSE;
+
 	if (ev_window->priv->document->iswebdocument == TRUE ) return ;
 	if (ev_window->priv->annot)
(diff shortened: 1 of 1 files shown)


raveit65:
You need to enable the sidebar and choose annotation from the dropdownmenu on top of the sidebar.
Than click on the button at the buttom of the sidebar and click again in the document.
That's it, the annotation window should popup at this point.

raveit65:
And a right-click on the annotation sign will show you 2 menuentries for editing and deleting the annotation.

lukefromdc:
All of that works, transparency included. Transparency applies to both the background and the foreground (text) color, not sure if that was intended or not. Annotations can now be removed as intended.

raveit65:
> Transparency applies to both the background and the foreground (text) color, not sure if that was intended or not.

Yes, text color is going transparent too. That's how it works with evince too. I will test if it is better not to make the text color transparent, if possible.

raveit65:
Yeah, in origin commit the deprecation is replaced.
But i wanted to use the same widgets like the other menuentries.
And as far as i know replacing Gtk_Action will remove our icons from menus.

PR #284, opened 2017-11-21
shell: forward accels to the focused widget

raveit65:
GtkWindow catches keybindings for the menu items _before_ passing them to
the focused widget. This is unfortunate and means that pressing ctrl+Left
arrow, Ctrl+Right arrow on the search bar ends up turning the EvView
instead of moving around the text.
Here we override GtkWindow's handler to do the same things that it
does, but in the opposite order and then we chain up to the grand
parent handler, skipping gtk_window_key_press_event.

See https://bugzilla.gnome.org/show_bug.cgi?id=676040

inspired from https://git.gnome.org/browse/evince/commit/?id=70a2c0780b1b44acfa18f4762a3400b89eb123b5

fix rhbz https://bugzilla.redhat.com/show_bug.cgi?id=1513826

Please test

### DIFF
--- a/shell/ev-window.c
+++ b/shell/ev-window.c
@@ -635,26 +635,34 @@ ev_window_update_actions (EvWindow *ev_window)
 	}
 }
 
+static const gchar *view_accels[] = {
+	"PageDown",
+	"PageUp",
+	"Plus",
+	"Minus",
+	"KpPlus",
+	"KpMinus",
+	"Equal",
+	"p",
+	"n",
+	"FitPage",
+	"FitWidth"
+};
+
 static void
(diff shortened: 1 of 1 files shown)


raveit65:
Yes it fixes using f, w, c and d with the searchbar.

raveit65:
I added a few more commits from upstream to remove unneeded things and to be stay more in sync with upstream.
@lukefromdc 
Please test again.

lukefromdc:
Seems to work fine with the latest changes

raveit65:
I got positive feedback by user at https://bugzilla.redhat.com/show_bug.cgi?id=1513826#c7

PR #303, opened 2018-03-20
some backports

raveit65:
- shell: Use gdbus-codegen for the org.mate.atril.Application interface
- shell: Use gdbus-codegen for the org.mate.atril.Window interface
- daemon: Use gdbus-codegen for the org.mate.atril.Daemon interface
- daemon: Port to GApplication
- shell: Use g_printerr instead of g_warning
- shell: Port to GtkApplication
- shell: Use GtkApplication to inhibit the screensaver
- thumbnailer: Send error messages to stderr
- shell: Remove unused marshalers

Please test.


lukefromdc:
Didn't see any warnings in terminal except those caused by document defects, and nothing in syslog either. Looks good on that front too.

monsta:
If https://github.com/mate-desktop/atril/commit/2f89fccf54a858495101cd57bac66ffca657ff62 works as good as the old "totem-screensaver" thing, we can do the same in eom, removing the other remaining copy of that.

raveit65:
Sure , this can be done.

PR #310, opened 2018-04-02
libview: fix text selection with mouse in annotation window

monsta:
taken from:
https://github.com/GNOME/evince/commit/91ac78bc587eb2ac931136c9da9609b142410209

fixes the same problem as we had in stickynotes applet: when you try to select the text with mouse, the window is dragged instead.

### DIFF
--- a/libview/ev-annotation-window.c
+++ b/libview/ev-annotation-window.c
@@ -280,12 +280,34 @@ ev_annotation_window_close (EvAnnotationWindow *window)
 	g_signal_emit (window, signals[CLOSED], 0);
 }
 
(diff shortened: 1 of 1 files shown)


raveit65:
confirmed, still wondering that i missed that.

raveit65:
I cherry-picked it to 1.20, i guess it doesn't hurt for 1.18.

monsta:
Yes, picked to 1.18 in https://github.com/mate-desktop/atril/commit/21254c434261d67fe6523e1b9615ac0ba9a54607.

PR #311, opened 2018-04-02
A11y fixes

raveit65:
Some more fixes from upstream.
Personal i couldn't reproduce what they fix because orca runs with or without them.
But i wouldn't be worse to use them.
@alexarnaud
Please test

### DIFF
--- a/libview/ev-page-accessible.c
+++ b/libview/ev-page-accessible.c
@@ -543,23 +543,26 @@ ev_page_accessible_get_range_for_boundary (AtkText          *text,
 	if (!log_attrs)
 		return;
 
+	if (offset < 0 || offset >= n_attrs)
+		return;
+
 	switch (boundary_type) {
 	case ATK_TEXT_BOUNDARY_CHAR:
 		start = offset;
 		end = offset + 1;
 		break;
 	case ATK_TEXT_BOUNDARY_WORD_START:
-		for (start = offset; start >= 0 && !log_attrs[start].is_word_start; start--);
-		for (end = offset + 1; end <= n_attrs && !log_attrs[end].is_word_start; end++);
+		for (start = offset; start > 0 && !log_attrs[start].is_word_start; start--);
+		for (end = offset + 1; end < n_attrs && !log_attrs[end].is_word_start; end++);
 		break;
 	case ATK_TEXT_BOUNDARY_SENTENCE_START:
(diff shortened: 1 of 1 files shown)


raveit65:
Click on the commits and you will find links to gnome bugzilla in commit body.
https://github.com/mate-desktop/atril/pull/311/commits/446890444f15b0b008e188ff827d64a1c09a0fc8
https://github.com/mate-desktop/atril/pull/311/commits/20aadc714eb3c3f5ad6aef7c219e2085cd3d8254

monsta:
Well, I can install Orca again... but I don't know how to test this with upstream's steps to reproduce.

> 3.  Press the key stroke to let Orca read the whole document, for laptop mode,
    that is capslock + arrow down.

Laptop mode? :confused:

raveit65:
Ok, it works for me, so i will merge it.

alexarnaud:
Le 05/04/2018 à 16:50, monsta a écrit :
> Well, I can install Orca again... but I don't know how to test this with 
> upstream's steps to reproduce.
> 
>      3. Press the key stroke to let Orca read the whole document, for
>         laptop mode,
>         that is capslock + arrow down.
> 
> Laptop mode? 😕

Laptop mode is the mode where Orca key is capslock, by default it is 
insert because by default it's a desktop mode.

If you can wait a bit I'll do the test, I was just in a meeting to 
present Hypra and I've found no time to do a proper test. To be honest, 
I trust 100% the upstream guys involved in the initial test. If the fix 
is exactly the same on Atril, there are no reasons to not work.

Best regards,
Alex.

monsta:
Thanks. It's merged by now, but of course some more testing won't hurt.

alexarnaud:
@raveit65 We've fixed  an accessibility issue in Evince here: https://gitlab.gnome.org/GNOME/evince/issues/889 Do you want us to submit it also to Atril ? I don't know exactly where to discuss that. We plan to fix in the next 2 years many issue on the accessibility implementation of Atril so it could be good to define the way you want us (Hypra) to collaborate with Atril.

As I remember there is no mailing list to discuss Mate devel.

Best regards,
Alex.

raveit65:
Please submit it to atril if you can.
If atril needs the same fixes than evince we can use discussion in evince as base.
Don't forget to add short steps to reproduce the issue and fix for it to PR.
This would help testing a lot.

PR #312, opened 2018-04-04
Basic backports

raveit65:
- backend: Remove pixbuf backend
Evince is a document viewer, not an image viewer. Eog does the latter job
much better than evince ever could

- thumbnailer: Remove unused code
The code to overlay an icon is unused since commit (2005)
https://git.gnome.org/browse/evince/commit/?id=808285c

Please test.

### DIFF
--- a/libdocument/ev-document-misc.c
+++ b/libdocument/ev-document-misc.c
@@ -229,68 +229,10 @@ ev_document_misc_pixbuf_from_surface (cairo_surface_t *surface)
 {
 	g_return_val_if_fail (surface, NULL);
 
-	GdkPixbuf       *pixbuf;
-	cairo_surface_t *image;
-	cairo_t         *cr;
-	gboolean         has_alpha;
-	gint             width, height;
-	cairo_format_t   surface_format;
-	gint             pixbuf_n_channels;
-	gint             pixbuf_rowstride;
-	guchar          *pixbuf_pixels;
-	gint             x, y;
-
-	width = cairo_image_surface_get_width (surface);
-	height = cairo_image_surface_get_height (surface);
-	
-	surface_format = cairo_image_surface_get_format (surface);
-	has_alpha = (surface_format == CAIRO_FORMAT_ARGB32);
-
-	pixbuf = gdk_pixbuf_new (GDK_COLORSPACE_RGB,
-				 TRUE, 8,
-				 width, height);
-	pixbuf_n_channels = gdk_pixbuf_get_n_channels (pixbuf);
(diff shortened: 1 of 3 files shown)


monsta:
I use Atril to view a large block diagram in jpg file, where I have to scroll in both directions and sometimes zoom in and out. I found Atril more convenient for this than eom.

raveit65:
I want to remove it because to avoid pain with HIDPI. As upstream removed it  we will find no help there.
Is there any chance to use another application or eom, eog for your jpg.....please?

raveit65:
And did it work with scale=2 for you?

raveit65:
@monsta 
Before you switch to xreader i will drop the commit.
Can you please review this one and https://github.com/mate-desktop/atril/pull/311 ?

raveit65:
@monsta 
Done, but i am sure this will cause problems later.
Any ways, i can disable building the backend for fedora.

monsta:
I can't test it with HiDPI. Are there any problems with that backend?
I think you don't have to disable it explicitly, it's disabled by default. Some distros might have it enabled, I know that Debian/Ubuntu/Mint enable it.

monsta:
BTW, we also have a TIFF backend here, which is also for images. Yet upstream didn't drop it...

raveit65:
Do you need the obsolete backend or not?
Again, this will cause issues with back porting commits.

raveit65:
> I can't test it with HiDPI. 

Why?
You only need to do the same like i do with a non HIDPI monitor.
Use windows-scaling=2 and you will see if it is sharp or blurred.

monsta:
I need that backend for now. I can try it with scaling when I'll get back to work, if it's possible to set the scaling for Atril only. I recall there were some environment variables for that, like GDK_SCALE or something?

lukefromdc:
Seems to work fine in a test here, and no issues with window-scaling=2. Late but this is still the development branch

monsta:
Thanks @vkareh :slightly_smiling_face: 

I've tested it with my large image, looks like the image is scaled more than 125%, can't say precisely, maybe it's even 200%. How should it be for HiDPI?

Found a problem as well. Even though the image is scaled that much, the zoom combobox shows 50%. Choosing other zoom levels doesn't actually change the zoom. This only affects combobox. Zooming with menu items or Ctrl+scroll works as before.

raveit65:
> Even though the image is scaled that much, the zoom combobox shows 50%. Choosing other zoom levels doesn't actually change the zoom.

Same as epubs, for this reason i want to remove a format which isn't supported by evince.

PR #313, opened 2018-04-04
Fix presentation actions

monsta:
To test the first commit:
- open truncated pdf (like [this one](https://bugzilla.redhat.com/show_bug.cgi?id=1536856#c20))
- press F5
- nothing should happen

Without the fix, you'll get to fullscreen (like with F11) and see some warnings in the terminal.

To test the second commit:
- open atril w/o documents or with truncated pdf
- press F11 to get to fullscreen
- Start Presentation button should be disabled

Without the fix, Start Presentation button is enabled. Clicking on it w/o documents makes atril crash. Clicking on it with truncated pdf leads to some warnings in the terminal.


lukefromdc:
These fixes both work. Sorry I could not test earlier but had my hands full elsewhere. Still needed testing though as this is still the development branch

PR #324, opened 2018-07-13
presentation: avoid deprecated 'gtk_style_context_set_background'

sc0w:
I see no changes in 'presentation' (F5) without deprecated code

### DIFF
--- a/libview/ev-view-presentation.c
+++ b/libview/ev-view-presentation.c
@@ -1313,9 +1313,6 @@ ev_view_presentation_realize (GtkWidget *widget)
 	gdk_window_set_user_data (window, widget);
 	gtk_widget_set_window (widget, window);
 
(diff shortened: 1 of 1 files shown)


raveit65:
At the end of a presentation there is a hint to close the presentation with `Esc`, so i think this is the intended behavior.

Can someone explain me why the code works with simply removing this function?

Beside from that i can confirm that the `w` (white) and `b` (black) keys to set presentation background are working with that change.

raveit65:
Maybe this code isn't needed anymore since we use a css for drawing the background?
https://github.com/mate-desktop/atril/commit/1d90b700c7bf5036e9f1e7ddd0adb09bd39467ce

sc0w:
@raveit65  

> Can someone explain me why the code works with simply removing this function?

sorry, I don't know

> Maybe this code isn't needed anymore since we use a css for drawing the background?

probably, I have no idea

PR #332, opened 2018-09-17
Update Catalan translation of manual

rbuj:


raveit65:
Translations will be done by transifex.
You need do it there.

rbuj:
The manual of Atril is not available in transifex.

raveit65:
Really...,
thanks for pointing that out , i will add the manual to transifex.

raveit65:
@rbuj 
I just added master--atril-user-quide and 1.20--atril-user-quide to transifex.
You can start with translations.
`tx pull -a -f` works fine for me.

rbuj:
Thanks. I sent the Catalan translation using:
$ tx push -l ca --translations --resource MATE.master--atril-user-quide

raveit65:
Nice, you can push a already translated .po file with tx-client to transifex?
... i didn't know that :)

PR #339, opened 2018-10-25
shell: Save document to the same path it was opened from

### LINKED ISSUE #322: Edited pdf saves to previously used directory, rather than directory in which pdf already exists
#### Expected behaviour
1. Open example1.pdf located at ~/Documents/example1/example1.pdf
2. Edit PDF
3. Click "save a copy"
4. File browser opens to directory ~/Documents/example1/example1.pdf. Click "save".
5. Open example2.pdf located at ~/Documents/example2/example2.pdf
6. Edit PDF
7. Click "save a copy"
8. File browser opens to directory ~/Documents/example2/example2.pdf

#### Actual behaviour

5. Open example2.pdf located at ~/Documents/example2/example2.pdf
6. Edit PDF
7. Click "save a copy"
8. File browser opens to directory ~/Documents/example1/example1.pdf

1. Open one document, make an edit, and save a copy.
2. Open another document from a different directory, make an edit, and click "save a copy"
3. The file browser will open to the location of the first document, rather than the location of the document currently being edited, resulting in inefficient workflow and misplaced documents.

#### MATE general version
1.20.1

#### Package version
1.20.1

#### Linux Distribution
Distributor ID:	Ubuntu
Description:	Ubuntu 18.04 LTS
Release:	18.04
Codename:	bionic

#### link to downstream report of your Distribution

vkareh:
When annotating or filling a form in a document, this must be saved
as a different document as atril does not overwrite documents.
The user can expect to store the modified file in the same place
than the original document, except when the document lives in
a temporary directory (e.g. downloaded automatically with a web
browser), in whose case it must fallback to the Documents
directory (if set) or the the home directory.

Previously, atril assumed the latest directory used, or the place
where an image or attachment was stored last. Such behaviour is
confusing because the latest place opened might have no relation
with the document modified.

Fixes https://bugzilla.gnome.org/show_bug.cgi?id=767611
Adapted from https://gitlab.gnome.org/GNOME/evince/commit/319a6d49

Fixes #322

### DIFF
--- a/shell/ev-window.c
+++ b/shell/ev-window.c
@@ -3207,11 +3207,12 @@ static void
 ev_window_cmd_save_as (GtkAction *action, EvWindow *ev_window)
 {
 	GtkWidget *fc;
-	gchar *base_name;
-	GFile *file;
+	gchar *base_name, *dir_name, *var_tmp_dir, *tmp_dir;
+	GFile *file, *parent;
+	const gchar *default_dir, *dest_dir, *documents_dir;
 
 	fc = gtk_file_chooser_dialog_new (
-		_("Save a Copy"),
+		_("Save As…"),
 		GTK_WINDOW (ev_window), GTK_FILE_CHOOSER_ACTION_SAVE,
 		"gtk-cancel", GTK_RESPONSE_CANCEL,
 		"gtk-save", GTK_RESPONSE_OK,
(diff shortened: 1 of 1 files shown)


raveit65:
This works fine, but we should rename button and text to `Save_As` in a dialog window, ihmo.
Same action name which is used in menu.
I closed atril with alt+f4 to trigger the dialog after editing the document.
atril-2018-10-25 18-00-27

raveit65:
Still wondering if there isn't a commit in evince for this :)

vkareh:
Good call, I can search evince, or I can just search/replace on my own. I'll take a look

Issue #341, opened 2018-11-05
Atril can't open cbr (rar) archives


terzag:
On Ubuntu MATE 18.10, Atril can't manage cbr (rar) archives anymore. I get an error message "RAR archive (application/vnd.rar) filetype is not supported" and thumbnails in Caja are not generated. Cbz (zip) archives work fine.

The issue actually already happened with Ubuntu 16.04 and was fixed in Atril 1.16.1. It seems to be back in 1.20.2.

I opened a bug report on [Launchpad](https://bugs.launchpad.net/ubuntu/+source/atril/+bug/1801704), not sure if it's more relevant there or here.

acastan:
The same happened to me with Atril and Evince in Debian testing last month. Stop opening cbr files but still opening cbz files.

Evince fixed it with this patch:

https://gitlab.gnome.org/GNOME/evince/issues/972

eli-schwartz:
This resolution is completely wrong. Atril does NOT support application/vnd.rar, which is the generic RAR format. It exclusively supports application/vnd.comicbook-rar, which is the subclass that specifically has the .cbr extension.

Failure to correctly follow the Freedesktop standard means that atril offers itself as a mimetype handler for all rar files whether they end in the Comic Book Archive format extension or not. Furthermore, the thumbnailer program will offer to thumbnail all RAR files (and file browsers may automatically do so), which is completely wrong.

Downstream bug report: https://bugs.archlinux.org/task/61128
This is resulting in users with very large RAR files that are not comic books, having those files be extracted for attempted thumbnailing.

PR #342, opened 2018-11-05
comics: support application/vnd.rar MIME type

sc0w:
Fixes https://github.com/mate-desktop/atril/issues/341

How to test:

See this sample **cbr** file without/with the PR

[testingcbr.zip](https://github.com/mate-desktop/atril/files/2550466/testingcbr.zip)

### DIFF
--- a/backend/comics/comicsdocument.atril-backend.in
+++ b/backend/comics/comicsdocument.atril-backend.in
@@ -1,4 +1,4 @@
 [Atril Backend]
 Module=comicsdocument
 _TypeDescription=Comic Books
-MimeType=application/x-cbr;application/x-cbz;application/x-cb7;application/x-cbt;application/vnd.comicbook+zip;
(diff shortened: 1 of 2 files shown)


selectiveduplicate:
The file opens fine with this PR's build:
``atril ~/Downloads/testing.cbr``

Although I do get the following two errors (aside from the usual ``atk-bridge`` errors):
```
** (atril:17282): CRITICAL **: 21:49:42.479: void webkit_web_view_load_uri(WebKitWebView*, const gchar*): assertion 'uri' failed

(atril:17282): GLib-CRITICAL **: 21:49:48.344: g_source_set_ready_time: assertion 'source->priv != NULL' failed
```
The last one occurs when I close atril.

lukefromdc:
I might have had a bad download of the test file, but the zip file opened and I was able to extract the file within it

sc0w:
@lukefromdc probably, I don't know, but it isn't related to this PR and it can be fixed in new PR.

I will merge this, it works fine.

Issue #346, opened 2018-12-14
Strange places found by PVS studio


shlyakpavel:
This is totally senseless. Please, inspect the code if original author has meant something else
https://github.com/mate-desktop/atril/blob/29fc232c5a7f28af4d3162d7ceebb65d9bd6b1e0/libdocument/ev-document-factory.c#L139
It is possible that break is missing here. Consider removing useless line otherwise
https://github.com/mate-desktop/atril/blob/29fc232c5a7f28af4d3162d7ceebb65d9bd6b1e0/backend/pdf/ev-poppler.cc#L703
It is also possible that is is not a good idea to do memset before checking tf_buf against null pointer here
https://github.com/mate-desktop/atril/blob/29fc232c5a7f28af4d3162d7ceebb65d9bd6b1e0/backend/tiff/tiff2ps.c#L1545
The expression is always false here
https://github.com/mate-desktop/atril/blob/29fc232c5a7f28af4d3162d7ceebb65d9bd6b1e0/shell/ev-message-area.c#L173
Pay attention to the if statement here
https://github.com/mate-desktop/atril/blob/29fc232c5a7f28af4d3162d7ceebb65d9bd6b1e0/shell/ev-window.c#L1564-L1566
It has identical sub-expressions surrounding || operator
Anyway, thank you guys for the great things you do!
Have a nice day :)

raveit65:
@mate-desktop/core-team

sc0w:
@shlyakpavel please, can you open new pull request with your suggested changes?

shlyakpavel:
@sc0w I don't think I have enough expertise to do that. These places should be checked by someone who understands the code logic.

monsta:
Thanks, we'll fix these. Two issues in backends are still in Evince too, so you can report them to Evince developers too if you wish.

PR #281, opened 2017-09-07
Implement caret navigation

raveit65:
This will enable caret navigation inside a pdf dokument and it works with orca.
The F7 key enables the caret-navigation.
I know this is a huge PR but the commit for enabling F7 was done very late at upstream.
And without the key you can't test something.
Mainly i want to know that i didn't broke any existing functions.
If something goes wrong i can split the PR.
At least this is a first step and i will port more commits from upstream.
This won't work with epubs. But i noticed that keyboard navigation was already enabled before with ebubs if you have  a text selection.
Please test.

PR #347, opened 2018-12-16
assorted fixes for issues found by PVS studio

monsta:
Fixes https://github.com/mate-desktop/atril/issues/346.

Most of this is straightforward fixing for code issues. The last commit is related to caret navigation, so I thought it would be better to give it some testing. It shouldn't crash on pressing F7 (see the discussion at https://github.com/mate-desktop/atril/pull/281 and https://github.com/mate-desktop/atril/commit/52283c0d4b0bb97a00570c2978fcfc5e3f697ddc). It also sets the a11y object name now, so @alexarnaud please test this too.

### DIFF
--- a/shell/ev-message-area.c
+++ b/shell/ev-message-area.c
@@ -140,7 +140,6 @@ ev_message_area_set_image_for_type (EvMessageArea *area,
 				    GtkMessageType type)
 {
 	const gchar *icon_name = NULL;
-	const gchar *stock_id = NULL;
 	AtkObject   *atk_obj;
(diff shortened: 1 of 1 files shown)


lukefromdc:
My test was with a PDF file, I don't have any .tiff files to test that backend with

monsta:
Oh great, we still have a copy of [synctex](http://www.tug.org/svn/texlive/trunk/Build/source/texk/web2c/synctexdir/) here. Evince also still has it.

PR #349, opened 2018-12-19
comics: fix incorrect mimetype which said to open and thumbnail all rar files

### LINKED ISSUE #341: Atril can't open cbr (rar) archives

eli-schwartz:
application/vnd.rar is for files with the .rar extension, application/vnd.comicbook-rar is the subtype for files with the .cbr extension.

This properly fixes #341

### DIFF
--- a/backend/comics/comicsdocument.atril-backend.in
+++ b/backend/comics/comicsdocument.atril-backend.in
@@ -1,4 +1,4 @@
 [Atril Backend]
 Module=comicsdocument
 _TypeDescription=Comic Books
-MimeType=application/x-cbr;application/x-cbz;application/x-cb7;application/x-cbt;application/vnd.comicbook+zip;application/vnd.rar;
+MimeType=application/x-cbr;application/x-cbz;application/x-cb7;application/x-cbt;application/vnd.comicbook+zip;application/vnd.comicbook-rar;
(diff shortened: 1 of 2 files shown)


lukefromdc:
Do we have a test file for this that is not a comics file? I need something to test a build of this against

eli-schwartz:
No, but I assume you can install the rebundled proprietary "rar" binary from the rarlabs website, on most distros (may have to enable the nonfree repository on debian-like systems or use the Arch Linux AUR), or just download the binary directly, and make one yourself.

If you have an existing CBR, renaming it to .rar should work, since the mimetype specification only differentiates between the two based on whether the filename glob is "*.cbr".

sc0w:
@eli-schwartz  please, can you modify the commit message to avoid the first split line?

eli-schwartz:
I squeezed off a few more characters.

It was under 80 characters anyway???

sc0w:
> It was under 80 characters anyway???

I don't know

Thanks!

PR #354, opened 2019-01-29
[ephy-zoom-control] Avoid deprecated g_type_class_add_private

yetist:
Please review and test it.


lukefromdc:
Code on this one is a simple change, and at least on my system it works fine

PR #356, opened 2019-01-30
Add pixbuf, tiff and xps TypeDescription to POTFILES.in

rbuj:


lukefromdc:
I literally don't know anything about working with the PO files, and I have unused locales purged to reduce root filesystem space for backup handling. I don't have poedit here, and if I try to test these with locales purged and no experience I would be likely to miss things. I will leave this for other team members who really do know about this portion of the packages to handle, as we really don't need problems getting into translations by poor testing. I don't feel competent to handle this.

Thanks though for posting a test procedure as it will likely save the testers some time

lukefromdc:
@raveit65 , @monsta , @yetist ? We need someone to have a look at this who knows what they are doing with the translations

yetist:
This patch seems to make the description of these three files translatable.

PR #391, opened 2019-07-29
document: Replace deprecated gdk_color_equal

raveit65:
gdk_color_equal is simple enough to replace it with code inline

taken from:
https://gitlab.gnome.org/GNOME/evince/commit/1944b86

Please review.


lukefromdc:
Used the wrong merge button, got a useless merge commit. Sorry about that, to fix it would require a force-push to master. Restored the branch in case folks decide to take things to that level.

PR #408, opened 2019-08-29
build: drop fallback check for webkit2gtk-3.0 and require webkit2gtk-4.0 >= 2.6.0

monsta:
webkit2gtk-3.0 is dead upstream for years


monsta:
I don't mind - but curious: why that version? From what I saw in Debian, 4.0 suffix appeared after 2.6.0...

rbuj:
@monsta You're right, WEBKIT_REQUIRED should be 2.6.0, from webkit ChangeLog:
```
2014-07-31  Carlos Garcia Campos  <e-mail>

        [GTK] Bump binary version for 2.6
        https://bugs.webkit.org/show_bug.cgi?id=133724

        Reviewed by Philippe Normand.

        * Source/cmake/OptionsGTK.cmake: Bump binary version to 4.0 and
        update library version numbers.
```

https://github.com/WebKit/webkit/blob/35c80c15fa37baafd72e651792e7f4d169bd18ee/ChangeLog-2018-01-01#L10010-L10018

monsta:
ok, done

PR #410, opened 2019-08-31
WIP: Travis CI: arch/fedora: make 'synctex' folder buildable

sc0w:
Before:

`SyncTeX.............:    system`

Later:

`SyncTeX.............:    internal`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -76,6 +76,7 @@ requires:
     - make
     - mate-common
     - poppler-glib
+    - kpathsea
     - webkit2gtk
     - which
     - yelp-tools
@@ -137,6 +138,7 @@ requires:
     - mate-common
     - poppler-glib-devel
     - redhat-rpm-config
+    - texlive-kpathsea
     - webkitgtk4-devel
     - yelp-tools
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -76,7 +76,6 @@ requires:
     - make
     - mate-common
     - poppler-glib
-    - texlive-bin
     - webkit2gtk
     - which
     - yelp-tools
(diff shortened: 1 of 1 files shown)


raveit65:
Why switching to internal `SyncTeX` for fedora?
My plan is to switch atril to use `SyncTeX` for fedora official builds.
So i would prefer to use the same for Travis CI.

sc0w:
@raveit65

just to build `synctex` folder and see the clang analyzer warnings in `atril.mate-desktop.dev`

sc0w:
but the dvi backend is disabled with this PR:

`    DVI Backend.........:    no`

closing by the moment...

raveit65:
Yes, i am building atril since the beginning with `texlive-lib-devel` for the DVI backend.
see https://src.fedoraproject.org/rpms/atril/blob/master/f/atril.spec#_51
```
# for the dvi back-end
%if 0%{?fedora} >= 24
BuildRequires:  texlive-lib-devel
%else
BuildRequires:  texlive-kpathsea-lib-devel
%endif
```

sc0w:
yes, but with `texlive-lib-devel` installed, `atril` doesn't compile with the internal `synctex` code

raveit65:
Yes, and i don't want compile atril-1.24 for fedora with the internal code any more.
Honestly, fedora and and archlinux are bleeding-edge distros. So it make sense to use SyncTeX from distro repos for travis CI, to catch possible changes from upstream SyncTeX.
Why not using a distros which use always older package versions for building with internal SyncTeX, eg. debian?

Before you break behaviour for fedora or Archlinux i would prefer to drop internal SyncTeX code, as you asked for in other PR.
I think the reason that evince keep the internal code is that evince support building for mirsosoft windows.
I don't think that someone wants to/or should build atril for win7/8/10 ;)

PR #411, opened 2019-08-31
remove 'synctex' internally, work with external library instead

sc0w:


oz123:
@raveit65 @lukefromdc removing this is really unhelpful if one does not use latex. You are forcing users to install latex related packages. Please make this optional.

PR #414, opened 2019-09-10
Added accelerator 't' for toggling visibility of the toolbar.

apalsson:
This adds an accelerator to toggle the toolbar just like the side panel.
"T" was chosen as an accelerator as it is not used by any other action and it also fits with the action target's name.

### DIFF
--- a/shell/ev-window.c
+++ b/shell/ev-window.c
@@ -6601,7 +6601,7 @@ static const GtkActionEntry entries[] = {
 /* Toggle items */
 static const GtkToggleActionEntry toggle_entries[] = {
 	/* View Menu */
-	{ "ViewToolbar", NULL, N_("_Toolbar"), NULL,
+	{ "ViewToolbar", NULL, N_("_Toolbar"), "t",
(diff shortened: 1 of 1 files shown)


vkareh:
@mate-desktop/core-team - is this something we're interested in? Seems like a reasonable addition, considering we already have single letter accelerator for many View settings in Atril

alexarnaud:
> @mate-desktop/core-team - is this something we're interested in? Seems like a reasonable addition, considering we already have single letter accelerator for many View settings in Atril

A question, what's happening if a user try to write the letter "T" inside a PDF form?

Best regards.

apalsson:
> A question, what's happening if a user try to write the letter "T" inside a PDF form?
> 
Nothing, focus on a edit widget inside a document disables accelerators.

alexarnaud:
Le 11/09/2019 à 08:04, A. Palsson a écrit :
> Nothing, focus on a edit widget inside a document disables accelerators.
> 

OK, thanks for the clarification.

Best regards.

PR #417, opened 2019-10-02
point info dialog website to github

thomas-mc-work:
For me as a user this is much more useful than the link to https://mate-desktop.org/, as it points me directly to the official repo instead of searching on the mate website – even without success.

### DIFF
--- a/shell/ev-window.c
+++ b/shell/ev-window.c
@@ -5448,7 +5448,7 @@ ev_window_cmd_help_about (GtkAction *action, EvWindow *ev_window)
 		"copyright", _("Copyright \xc2\xa9 1996–2009 The Evince authors\n"
 		               "Copyright \xc2\xa9 2012–2019 The MATE developers"),
 		"license", license_trans,
-		"website", "https://mate-desktop.org/",
+		"website", "https://github.com/mate-desktop/atril",
 		"comments", comments,
 		"authors", authors,
 		"documenters", documenters,


raveit65:
All our apps use our website as link in info dialog, so this change doesn't make any sense.

vkareh:
This change would need to be done for all repositories. Ideally, our mate-desktop.org website would have a section for each program that we can link to, like mate-desktop.org/atril, and in there we should put a description to what Atril is, and links for file bugs and stuff.

Until such a page is created, we're better off leaving it as is. Just my 2 cents. Thoughts @mate-desktop/core-team ?

selectiveduplicate:
> This change would need to be done for all repositories. Ideally, our mate-desktop.org website would have a section for each program that we can link to, like mate-desktop.org/atril, and in there we should put a description to what Atril is, and links for file bugs and stuff.
> 
> Until such a page is created, we're better off leaving it as is. Just my 2 cents. Thoughts @mate-desktop/core-team ?

I agree with @vkareh. We already have a small description of the core applications on our website's homepage. Those can be revamped, and is a far better choice; but this doesn't make sense to me.

We can put a link to the atril repo there on the website if we want to.

flexiondotorg:
I agree, we should update the mate-desktop.org website to sign post where the source for each component is located. I'll take that on.

thomas-mc-work:
What has happened here?

raveit65:
No one from team like the change...

raveit65:
anyway, reopen ......

lukefromdc:
Any updates on this?

PR #419, opened 2019-10-14
Fix segfault on empty navMap

errt:
Port of https://github.com/linuxmint/xreader/pull/334 to fix https://github.com/linuxmint/xreader/issues/274 that also affects atril.

### DIFF
--- a/backend/epub/epub-document.c
+++ b/backend/epub/epub-document.c
@@ -1267,7 +1267,7 @@ setup_document_index(EpubDocument *epub_document,gchar *containeruri)
 
     xmlNodePtr navPoint = xmlretval;
 
-    do {
+    while(navPoint != NULL) {
 
         if ( !xmlStrcmp(navPoint->name,(xmlChar*)"navPoint")) {
     		xmlretval = NULL;
@@ -1315,7 +1315,7 @@ setup_document_index(EpubDocument *epub_document,gchar *containeruri)
 
         navPoint = navPoint->next;
 
-    } while(navPoint != NULL);
+    }
 
 	xml_free_doc();
 


raveit65:
Example document for reproducing the crash, please?

raveit65:
@errt 
Example document for reproducing the crash, please?

raveit65:
@errt 
We can't add code when author doesn't answer.
...but we can close.

errt:
Sorry, I'm not a user of atril, but of xreader and only reported this here, too, as it uses the same code base. As xreader has merged the proposed fix, I did not keep track of this anymore. I don't think I have a document at hand that triggers the reported behavior and I'm not sure I could craft a minimal example. As I personally don't care that much about atril and this is not a major bug, feel free to close it if you want to. On the other hand, I don't see much harm in the proposed change, it just ensures that no error can occur if a document, in violation of the standard, has an empty navMap. Your choice whether you consider this useful or not.

raveit65:
@errt 
I asked several times for a example document to reproduce the issue. Can you provide one?
I do not have such a document for myself.

rbuj:
Test document: guide2.epub (without navMap) and guide2.epub (with navMap)
```
wget http://www.ibiblio.org/pub/Linux/docs/LDP/users-guide/user-beta-1-src.tar.gz
tar xzvf user-beta-1-src.tar.gz 
cd user
pandoc -s guide.tex -o guide.epub
mkdir -p epub
unzip guide.epub -d epub
```
delete navMap:
```
$ grep -nR navMap
epub/EPUB/toc.ncx:12:  <navMap>
epub/EPUB/toc.ncx:1117:  </navMap>
$ sed -i 12,1117d epub/EPUB/toc.ncx
```
compress:
```
$ (cd epub && zip -rX ../guide2.epub  mimetype META-INF EPUB)
$ gzip guide2.epub 
$ gzip guide.epub
```
[guide.epub.gz](https://github.com/mate-desktop/atril/files/4902573/guide.epub.gz)
[guide2.epub.gz](https://github.com/mate-desktop/atril/files/4902575/guide2.epub.gz)

errt:
The missing navMap may be a different, but possibly related issue.
An empty navMap crashes atril 1.24 here, but does not crash xreader 2.6.4. Maybe a change since atril 1.24 already mitigated the issue.

PR #421, opened 2019-11-05
Gettext

yetist:
Note: sq.po has error.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -198,10 +198,6 @@ before_scripts:
   - fi
 
 build_scripts:
-  - if [ ${DISTRO_NAME} == "debian" ];then
-  -     curl -Ls -o debian.sh https://github.com/mate-desktop/mate-dev-scripts/raw/master/travis/debian.sh
-  -     bash ./debian.sh
-  - fi
   - ./autogen.sh
   - scan-build $CHECKERS ./configure
   - if [ $CPU_COUNT -gt 1 ]; then
--- a/.travis.yml
+++ b/.travis.yml
@@ -70,7 +70,6 @@ requires:
     - gcc
     - git
     - gobject-introspection
-    - intltool
     - itstool
     - libgxps
     - make
@@ -84,12 +83,12 @@ requires:
   debian:
     # Useful URL: https://github.com/mate-desktop/debian-packages
     # Useful URL: https://salsa.debian.org/debian-mate-team/atril
+    - autopoint
     - clang
     - clang-tools
     - cppcheck
     - git
     - gobject-introspection
-    - intltool
     - libcaja-extension-dev
     - libdjvulibre-dev
     - libgail-3-dev
@@ -144,11 +143,11 @@ requires:
     - yelp-tools
 
   ubuntu:
+    - autopoint
     - clang
     - clang-tools
     - git
     - gobject-introspection
-    - intltool
     - libcaja-extension-dev
     - libdjvulibre-dev
     - libgail-3-dev
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,6 +1,7 @@
 ACLOCAL_AMFLAGS = -I m4 ${ACLOCAL_FLAGS}
 
 SUBDIRS =       \
+    po          \
     cut-n-paste \
     data        \
     libdocument \
@@ -9,7 +10,6 @@ SUBDIRS =       \
     libmisc     \
     properties  \
     shell       \
-    po          \
     help
 
 if ENABLE_TESTS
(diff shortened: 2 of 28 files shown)


sc0w:
I requested to join Albanian team at transifex.

This is the fix:

```patch
--- a/po/sq.po
+++ b/po/sq.po
@@ -1109,7 +1109,7 @@ msgstr "Gërmat"
 #: ../shell/ev-properties-fonts.c:163
 #, c-format
 msgid "Gathering font information… %3d%%"
-msgstr "Mbledhja e informacionit font...%3d%%%"
+msgstr "Mbledhja e informacionit font...%3d%%"
 
 #: ../shell/ev-properties-license.c:134
 msgid "Usage terms"

**Comment by raveit65** [comment]
@sc0w 
I added you to project maintainers at transifex.
You should be able to edit every language now.

**Comment by sc0w** [comment]
@raveit65 ok, thanks! I fixed it at transifex

I did `tx pull -af` (the build is ok)

and I see the change `"Project-Id-Version: PACKAGE VERSION\n"` instead `"Project-Id-Version: MATE Desktop Environment\n"` in a lot of languages, do you know why?

**Comment by raveit65** [comment]
> and I see the change "Project-Id-Version: PACKAGE VERSION\n" instead "Project-Id-Version: MATE Desktop Environment\n" in a lot of languages, do you know why?

Not for shure, but i think this happens because `MATE Desktop Environment` isn't package name.
First i fixed that with a sed command before i pushed `a pull from transifex` to github the last 2 years.
But i after a while i ignore it because it doesn't break building or distcheck and it comes back every time after a pull from transifex.
This problem will gone after migration to gettext. See how it looks https://github.com/mate-desktop/mate-desktop/blob/master/po/af.po#L12

You shound only take care if `Content-Type: text/plain; charset=UTF-8\n` is correct after a pull from transifex. Than commit it to master and rebase this PR against master, please.

**Comment by sc0w** [comment]
ah, I know why, since https://github.com/mate-desktop/atril/commit/d3fd6379568c57e63eae54db044712b475b4e511 in `atril.pot` it was changed to `PACKAGE VERSION`

we can change it with `sed`, inside `makepot`

**Comment by raveit65** [comment]
> ah, I know why, since https://github.com/mate-desktop/atril/commit/d3fd6379568c57e63eae54db044712b475b4e511 in `atril.pot` it was changed to `PACKAGE VERSION`

`PACKAGE VERSION` was from pull from transifex. I didn't fix it before commiting because it doesn't break anything and it comes repeating back with every pull from transifex,
So i decided not to spend my time with it again, again and again...... ;)

> we can change it with `sed`, inside `makepot`

It looks OK after getext port or not?
```
"Project-Id-Version: atril 1.23.1\n"
```
In general i think using `atril ` as project name for a single repo is correct, and using the parent project name `MATE Desktop Environment` was already wrong in the past.
And i want to avoid to change that with `sed` command 2 times.
First after generating the `.pot` resource with `makepot` and second in `.mo` files after every pull from transifex.

**Comment by sc0w** [comment]
@raveit65 

> It looks OK after getext port or not?

"Project-Id-Version: atril 1.23.1\n"

yes, understood, thanks for the clarifications.

**Comment by sc0w** [comment]
@yetist we need the debian patch for `intltool` here?

I mean:
```
  - if [ ${DISTRO_NAME} == "debian" ];then
  -     curl -Ls -o debian.sh https://github.com/mate-desktop/mate-dev-scripts/raw/master/travis/debian.sh
  -     bash ./debian.sh
  - fi
```

yetist:
> @yetist we need the debian patch for `intltool` here?
> 

After migrating to gettext, we no longer need this patch, I will do it.

raveit65:
@yetist 
There are desktop files without translator hints :)
eg.
data/atril.desktop.in.in
properties/libatril-properties-page.caja-extension.desktop.in.in

The other desktop files for backends don't `icon Keywords or Categories`

raveit65:
@sc0w 
Should i merge it or do you want to approve it?

sc0w:
the macro `AM_GNU_GETTEXT_REQUIRE_VERSION` is available since `gettext 0.19.6`

for this reason we don't need `AM_GNU_GETTEXT_VERSION` because we require 0.19.8 in this PR.

Reference: https://lists.gnu.org/archive/html/info-gnu/2015-09/msg00003.html

> A new macro AM_GNU_GETTEXT_REQUIRE_VERSION can be used to indicate autopoint to pull the latest available infrastructure, instead of the exact version specified with AM_GNU_GETTEXT_VERSION.  When AM_GNU_GETTEXT_REQUIRE_VERSION is used, AM_GNU_GETTEXT_VERSION is ignored.

sc0w:
forget my comment, seems `autoreconf` needs `AM_GNU_GETTEXT_VERSION`

PR #450, opened 2020-04-13
Remove strict-prototypes warning

rbuj:
```
epub-document.c:438:1: warning: function declaration isn’t a prototype [-Wstrict-prototypes]
  438 | xml_free_doc();
      | ^~~~~~~~~~~~
epub-document.c:575:1: warning: function declaration isn’t a prototype [-Wstrict-prototypes]
  575 | xml_free_doc()
      | ^~~~~~~~~~~~
--
fontmap.c:648:13: warning: function declaration isn’t a prototype [-Wstrict-prototypes]
  648 | static void init_static_encoding()
      |             ^~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/backend/dvi/mdvi-lib/fontmap.c
+++ b/backend/dvi/mdvi-lib/fontmap.c
@@ -645,7 +645,7 @@ void	mdvi_install_fontmap(DviFontMapEnt *head)
 	}
 }
 
-static void init_static_encoding()
+static void init_static_encoding(void)
 {
 	DviEncoding	*encoding;
 	int	i;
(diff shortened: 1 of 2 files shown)


backend/dvi/mdvi-lib/fontmap.c
```diff
@@ -645,7 +645,8 @@ void	mdvi_install_fontmap(DviFontMapEnt *head)
 	}
 }
 
-static void init_static_encoding()
+static void
+init_static_encoding (void)
```
sc0w, line 649: this change breaks the style of the whole file, I think you can do the job in one line

rbuj: I agree, because it's an external code. http://mdvi.sourceforge.net/

PR #451, opened 2020-04-15
Fix cppcheck [knownConditionTrueFalse] warnings

sc0w:
Fixes the warnings:

```
backend/comics/comics-document.c:767:21: style: Condition 'bytes<=0' is always true [knownConditionTrueFalse]

backend/epub/minizip/unzip.c:603:25: style: Condition 'unz_copyright[0]!=' '' is always false [knownConditionTrueFalse]

properties/ev-properties-view.c:197:33: style: Condition 'text' is always true [knownConditionTrueFalse]

shell/ev-sidebar-thumbnails.c:598:7: style: Condition '!loading_icon' is always true [knownConditionTrueFalse]

shell/main.c:115:6: style: Condition '!error' is always true [knownConditionTrueFalse]
```

### DIFF
--- a/backend/epub/minizip/unzip.c
+++ b/backend/epub/minizip/unzip.c
@@ -119,9 +119,6 @@
 #define SIZEZIPLOCALHEADER (0x1e)
 
 
-const char unz_copyright[] =
-   " unzip 1.01 Copyright 1998-2004 Gilles Vollant - http://www.winimage.com/zLibDll";
-
 /* unz_file_info_interntal contain internal info about a file in zipfile*/
 typedef struct unz_file_info64_internal_s
 {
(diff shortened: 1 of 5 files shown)


backend/epub/minizip/unzip.c
```diff
@@ -600,9 +600,6 @@ local unzFile unzOpenInternal (const void *path,
 
     int err=UNZ_OK;
 
-    if (unz_copyright[0]!=' ')
```
rbuj, line 603: You can also remove `unz_copyright` vble.
https://github.com/mate-desktop/atril/blob/4852c2a1ec70597be303de82d01df68a54e3a44d/backend/epub/minizip/unzip.c#L122-L123

sc0w: thanks, done

PR #436, opened 2020-02-07
Make synctex an optional dependency

oz123:
This is still work in progress. The code still needs to be disabled
in the C code.
Comments are welcome, I am a total idiot when it comes to autotools :-P.

PR #464, opened 2020-05-28
build: Make synctex optional (--disable-synctex)

### LINKED ISSUE #438: [Compiling atril from source] Would it be possible to make synctex optional?
Hello mate-desktop devs,

I am trying to compile atril 1.24.0 right now.

The configure line shows this error:

    checking for SYNCTEX... no
    configure: error: "SyncTeX support is disabled since library version 1.21 or newer not found"

I looked at configure --help output but could not find a way to disable this.

Would it be possible to make synctex optional? I would be fine with reduced functionality, e. g. just as a simple .pdf file. I may have missed it but right now there appears to be no option in ./configure to allow that (I grepped for "sync" and "tex" but found no results).

rbuj:
Closes #438

### DIFF
--- a/backend/dvi/dvi-document.c
+++ b/backend/dvi/dvi-document.c
@@ -229,11 +229,15 @@ dvi_document_finalize (GObject *object)
 	G_OBJECT_CLASS (dvi_document_parent_class)->finalize (object);
 }
 
+#ifdef HAVE_SYNCTEX
+
 static gboolean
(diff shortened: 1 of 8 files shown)


sc0w:
https://github.com/mate-desktop/atril/pull/436 was already opened before, did you test/review it?

rbuj:
> #436 was already opened before, did you test/review it?

I didn't see it because it has no reference to the issue.

PR #471, opened 2020-06-29
ev-properties-dialog: Add on_notebook_scroll_event

rbuj:
Based on:
- https://github.com/mate-desktop/mate-terminal/commit/8d2eb0898154f2624e2ae34039169f307a3411a5
- https://github.com/mate-desktop/mate-terminal/commit/622e2d1272bf2d7a99d147482ae6a51b0a77f188

### DIFF
--- a/shell/ev-properties-dialog.c
+++ b/shell/ev-properties-dialog.c
@@ -52,6 +52,64 @@ ev_properties_dialog_class_init (EvPropertiesDialogClass *properties_class)
 {
 }
 
+static gboolean
+on_notebook_scroll_event (GtkWidget        *widget,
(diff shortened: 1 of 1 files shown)


shell/ev-properties-dialog.c
```diff
@@ -70,6 +128,12 @@ ev_properties_dialog_init (EvPropertiesDialog *properties)
 			                 GTK_RESPONSE_CANCEL);
 
 	properties->notebook = gtk_notebook_new ();
```
raveit65, line 130: Indents...

rbuj: done

PR #488, opened 2020-08-27
Typo in fr.po

ymollard:
Just updating a typo in `fr.po`. Is mate using transiflex? I can't find mate there, only the teams.


raveit65:
Yes, we are using transifex.
https://www.transifex.com/mate/MATE/dashboard/

ymollard:
OK thank you

PR #503, opened 2020-12-02
build: remove legacy variable MATEICONDIR

rbuj:


raveit65 (approved):
LGTM,
/usr/share/pixmaps isn't used.

PR #531, opened 2021-06-04
ev-window.c: only create a EvWebView if needed

### LINKED ISSUE #529: Unnecessary WebKit Processes
#### Expected behaviour
As I understand it, the WebKit dependency in Atril is only required for reading e-books (epub format). Regular PDF viewing functionality does not depend on WebKit. Therefore, WebKit processes should only be created when the user actually requests to open a file type requiring WebKit to be processed.

#### Actual behaviour
For each PDF document which is opened, Atril creates two WebKit processes, consuming over 80 MiB of memory:
```
linuxr      5414  0.2  2.5 103115832 100528 ?    Sl   12:12   0:12          |   \_ atril /home/linuxr/Desktop/document.pdf
linuxr      5433  0.0  1.1 102551456 45744 ?     SLl  12:12   0:00          |       \_ /usr/lib/aarch64-linux-gnu/webkit2gtk-4.0/WebKitWebProcess 7 28
linuxr      5434  0.0  0.9 85766032 38676 ?      SLl  12:12   0:00          |       \_ /usr/lib/aarch64-linux-gnu/webkit2gtk-4.0/WebKitNetworkProcess 8 28
```
This seems just a waste of resources and is also bad from a security perspective.

Open any PDF document with Atril and inspect the created OS processes.

#### MATE general version
1.24.0

#### Package version
1.24.0

#### Linux Distribution
Ubuntu 20.04.2 LTS

mbkma:
A WebKit processes should only be created when the user actually requests to open a file type requiring WebKit to be processed.

Fixes #529

### DIFF
--- a/shell/ev-window.c
+++ b/shell/ev-window.c
@@ -1787,6 +1787,14 @@ ev_window_set_document (EvWindow *ev_window, EvDocument *document)
 		ev_view_disconnect_handlers(EV_VIEW(ev_window->priv->view));
 		g_object_unref(ev_window->priv->view);
 		ev_window->priv->view = NULL;
+
+		if (ev_window->priv->webview == NULL)
+		{
+			ev_window->priv->webview = ev_web_view_new();
+			ev_web_view_set_model(EV_WEB_VIEW(ev_window->priv->webview),ev_window->priv->model);
+			ev_web_view_reload(EV_WEB_VIEW(ev_window->priv->webview));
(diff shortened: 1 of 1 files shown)


raveit65 (approved):
LGTM,
webkit process aren't when only a pdf or other document is open.
Also the webkit process will be clean terminated when you close a epub but another pdf is still open.

PR #541, opened 2021-10-24
Use GLib's new g_clear_signal_handler() function to simplify code

rbuj:


raveit65:
@rbuj
Can you please push PR to get a deployment?

rbuj:
done

lukefromdc (approved):
I only use a limited set of features with atril and the PDF documents I have so could not test everything (notably could not test annotations) but in basic interaction witn PDFs behavior seems to be the same as in master. Zoom control (affected by this) works fine with no new issues on my end

raveit65:
@rbuj 
Deployment is missing.

raveit65:
In a quick test it seems ok, i'd like to test it a few days more.

raveit65 (approved):
LGTM

PR #544, opened 2021-10-26
Preserve compatibility with older GLib versions

vvillenave:
This is a rather non-invasive change, that will  
be easy to remove whenever we go ahead and  
stop supporting older systems.


raveit65 (approved):
LGTM

Issue #325, opened 2018-07-22
EPUB fails to open


Carmina16:
#### Actual behaviour
This EPUB file from an online library does not open itself.
[testatril_mib.zip](https://github.com/mate-desktop/atril/files/2217176/testatril_mib.zip)

#### MATE general version
N/A

#### Package version
1.20

#### Linux Distribution
Alt Sisyphus

rbrito:
I have seen **a lot** of epub files that I can't open with atril (actually, they only show a blank page and I can see no way to navigate them).

OTOH, with very simple files, like the one that I describe at <https://github.com/rbrito/epub-example>, I can see the contents of the file.

lukefromdc:
Are any of those files DRM ("digital restrictions management") encumbered?

selectiveduplicate:
Trying to open the file that @Carmina16 has given, I get the following error message:

**File type Zip archive (application/zip) is not supported**

It's a .epub file, but it gets recognized as a Zip archive?

Carmina16:
It is a ZIP file containing the EPUB in question, so you need to extract it first. It has no DRM and if a valid ZIP.

sc0w:
I can open the epub mentioned with fbreader 0.12.10 but atril fails

rcaridade145:
https://github.com/mate-desktop/atril/blob/master/backend/epub/epub-document.c#L621 

` gchar* mimetypes[] = {"application/epub+zip","application/x-booki+zip"};`

https://www.iana.org/assignments/media-types/application/epub+zip

tsjnachos117:
I could be wrong, but I think \***ALL**\* epub books are essentially a bunch of files (probably mostly *.html, *,xml, and standard images such as *.jpg) combined together in a zip file. You could modify an epub by extracting it, changing individual files, and re-zipping it.

Anyway, I wanted to point out that this is probably related to linuxmint/xreader#266.

rbrito:
@tsjnachos117, indeed, as I describe in my toy [epub example][0], an epub file is a zip file. I also wrote a [simple script][1] to optimize the size of an epub (especially useful for saving size in mobile devices).

[0]: https://github.com/rbrito/epub-example
[1]: https://github.com/rbrito/scripts/commits/master/repack-epub.sh

rbrito:
Just as an extra point, the original epub posted here doesn't have the mimetype file as its first file, which, I believe, makes it non-compliant.

[...]

PR #555, opened 2021-12-23
Trust the file extension for EPUBs, fixes mate-desktop/atril#325

fogbank:
Atril was doing a "deep" (the first 1024 bytes) MIME type guessing and this didn't work with some EPUB files.

Aside from the fact that GLib seems to be able to do a better job analyzing the file's URI on its own (are there any contraindications to this?), I preferred to simply do away with the deep check, since there's another MIME type check involved with EPUBs that executes before this one, one that checks for the presence of WebKit, and that one trusts the system's MIME type database (which in my case is based on the file's extension.)

Oh, and I have also added the VS Code config dir to the .gitignore make target because that's what I use. I hope I've done it the right way; I've been a programmer for a long time, but I'm VERY new w.r.t. creating pull requests and collaborating on GitHub (or in general), and I haven't touched C since before I turned pro (that's more that two decades ago.)

### DIFF
--- a/git.mk
+++ b/git.mk
@@ -367,6 +367,7 @@ $(srcdir)/.gitignore: Makefile.am $(top_srcdir)/git.mk $(top_srcdir)/configure.a
 			"*~" \
 			".*.sw[nop]" \
 			".dirstamp" \
+			".vscode/" \
 		; do echo "/$$x"; done; \
 		for x in \
 			"*.$(OBJEXT)" \
--- a/backend/epub/epub-document.c
+++ b/backend/epub/epub-document.c
@@ -624,7 +624,7 @@ static gboolean
 check_mime_type(const gchar* uri,GError** error)
 {
     GError * err = NULL ;
-    const gchar* mimeFromFile = ev_file_get_mime_type(uri,FALSE,&err);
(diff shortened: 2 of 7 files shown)


raveit65:
This was a bad idea to commits from other PRs and squash them.
Your branch only need to be in sync with master.
Sorry, we can't test and merge your like this, please revert last commit.

git.mk
```diff
@@ -367,6 +367,7 @@ $(srcdir)/.gitignore: Makefile.am $(top_srcdir)/git.mk $(top_srcdir)/configure.a
 			"*~" \
 			".*.sw[nop]" \
 			".dirstamp" \
+			".vscode/" \
```
raveit65, line 370: For what is that visual studio code entry good?

vkareh:
@fogbank can you create a new branch on your repo, then cherry-pick the actual fix commit into it, then create a new PR with it? This PR as-is is problematic.

git.mk
vkareh: @fogbank please remove this entry. You can configure a global ~/.gitconfig to deal with any special configuration in your machine.

raveit65:
No answer from author, i opened a nw PR with changes form here.
https://github.com/mate-desktop/atril/pull/570
Closing this.

PR #571, opened 2022-08-07
epub: add fallback for malformed epub files in check_mime_type

### LINKED ISSUE #325: EPUB fails to open

rbuj:
fix #325

### DIFF
--- a/backend/epub/epub-document.c
+++ b/backend/epub/epub-document.c
@@ -623,41 +623,51 @@ xml_get_data_from_node(xmlNodePtr node,
 static gboolean
 check_mime_type(const gchar* uri,GError** error)
 {
-    GError * err = NULL ;
-    const gchar* mimeFromFile = ev_file_get_mime_type(uri,FALSE,&err);
+    GError * err = NULL;
+    const gchar* mimeFromFile;
 
-    gchar* mimetypes[] = {"application/epub+zip","application/x-booki+zip"};
-    int typecount = 2;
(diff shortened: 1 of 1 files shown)


lukefromdc (approved):
All the build warnings I get from compiling epub-document.c with this applied are from parts of the code which are NOT changed and shared with master, so looks like no new build warnings. A test epub file opens fine with this installed,and no runtime warnings show up in terminal.

I have not had enough experience with epub files to run into the original issue, but this code does seem to work with a known good test file. Security-wise, refusing to open malformed files can block an attack on someone's machine. Downside is it may take a bit less corruption to make a file impossible to open.

Probably safer than trusting a file extension which a malicious document author can easily slap on to anything.

lukefromdc:
A rebase of this against current master opened the previously referrred to test file from the issue, though being malformed internal elements of it beyond the picture were not found. They were not found with just trusting the file extension either, and that's potentially less safe, so that's the fault of the file, not this effort to get partially usable epub files to open.

 Another test file 
https://github.com/IDPF/epub3-samples/releases/download/20170606/cc-shared-culture.epub
from the older segfault issue fixed by another PR opened fine with this applied.

PR #591, opened 2023-09-30
Remove compilation warnings and set 'always_show_image' Properties in the UI file

zhuyaliang:


raveit65:
Commit works fine. I think you can delete double `data/atril-toolbar.xml` and `data/atril-ui.xml`.
They are already /shell dir and not in extradist of /data dir.
So they was never in released tarball and would use for building.

lukefromdc (approved):
This suppressed 4 build warnings with my normal build options (there are many others at this time). The build finished with no issues and runtime behavior was same as before

zhuyaliang:
If that's the case, we can delete this file in the next submission

raveit65:
> If that's the case, we can delete this file in the next submission

Good, i have already a commit prepared for this.

raveit65 (approved):
LGTM

PR #592, opened 2023-10-01
drop double ui files

raveit65:
They are already in /shell dir.
In /data dir they're not in extradist, in result they  are not in released tarballs and won't build.
Local i created a new release tarball and tested them with this commit. Works fine without surprises.


zhuyaliang (approved):
Compilation is normal

lukefromdc (approved):
Works fine, no problems with this building or at runtime

PR #597, opened 2023-12-13
Fix build with libxml2 2.12

jtojnar:
libxml 2.12.0 reorganized headers, resulting in xmlParseFile and xmlIndentTreeOutput no longer being in scope.
Let’s add the proper includes containing the symbols.

### DIFF
--- a/cut-n-paste/toolbar-editor/egg-toolbars-model.c
+++ b/cut-n-paste/toolbar-editor/egg-toolbars-model.c
@@ -27,7 +27,9 @@
 
 #include <unistd.h>
 #include <string.h>
+#include <libxml/parser.h>
 #include <libxml/tree.h>
+#include <libxml/xmlsave.h>
 #include <gdk/gdk.h>
 
 static void egg_toolbars_model_finalize   (GObject               *object);


lukefromdc (approved):
Builds fine here, like the rest of these. Again though, I've still got libXML 2.9 even on Debian Unstable so not tested with the newest version.

jtojnar:
2.12.3 returned the cyclic include between `xml/tree.h` and `xml/parser.h` so it is no longer strictly necessary but I still think it is a good practice to include headers for required symbols explicitly.

raveit65:
Where does a build not running?
I did a test build with atril-1.26.1 at rawhide branch at fedora build server and i do not get an build error.
Fedora rawhide use [libxml2-2.12.3](https://koji.fedoraproject.org/koji/buildinfo?buildID=2332733)
successful build:
https://koji.fedoraproject.org/koji/taskinfo?taskID=111154090
I recall that i fixed all repos of us building with  libxml2 2.12  a while ago, and atril didn't fail.

jtojnar:
This was atril 1.26.1 on NixOS with libxml2 2.12.2, IIRC. Specifically, `xmlParseFile` and `xmlIndentTreeOutput` symbols were missing because `<libxml/tree.h>` no longer included `<libxml/parser.h>`.

The build you linked passes since libxml2 2.12.3 it contains re-added the cyclic import.

raveit65:
I tested all repos with  building against libxml2 2.12 a while ago, and atril didn't fail.
So, i am still wondering about your build issue with that version.

raveit65 (approved):
Not needed but it doesn't hurt.

Issue #603, opened 2024-01-24
gcc14: Latest 1.26.2 release failed to build on fedora rawhide


raveit65:
#### Expected behaviour

Succesful build on fedora rawhide
#### Actual behaviour
Latest 1.26.2 release failed to build on fedora rawhide

#### Steps to reproduce the behaviour
Latest 1.26.2 release failed to build on fedora rawhide.
from https://kojipkgs.fedoraproject.org//work/tasks/8848/112258848/build.log
```
epub-document.c: In function 'check_mime_type':
epub-document.c:635:40: error: passing argument 1 of 'g_strv_length' from incompatible pointer type [-Wincompatible-pointer-types]
  635 |         for (i = 0; i < g_strv_length (mimetypes); i++) {
      |                                        ^~~~~~~~~
      |                                        |
      |                                        const gchar ** {aka const char **}
In file included from /usr/include/glib-2.0/glib/gstring.h:37,
                 from /usr/include/glib-2.0/glib/giochannel.h:36,
                 from /usr/include/glib-2.0/glib.h:56,
                 from /usr/include/glib-2.0/glib/gi18n-lib.h:23:
/usr/include/glib-2.0/glib/gstrfuncs.h:383:55: note: expected 'gchar **' {aka 'char **'} but argument is of type 'const gchar **' {aka 'const char **'}
  383 | guint                 g_strv_length    (gchar       **str_array);
      |                                         ~~~~~~~~~~~~~~^~~~~~~~~
epub-document.c:646:48: error: passing argument 1 of 'g_strv_length' from incompatible pointer type [-Wincompatible-pointer-types]
  646 |                 for (i = 0; i < g_strv_length (mimetypes); i++) {
      |                                                ^~~~~~~~~
      |                                                |
      |                                                const gchar ** {aka const char **}
/usr/include/glib-2.0/glib/gstrfuncs.h:383:55: note: expected 'gchar **' {aka 'char **'} but argument is of type 'const gchar **' {aka 'const char **'}
  383 | guint                 g_strv_length    (gchar       **str_array);
      |                                         ~~~~~~~~~~~~~~^~~~~~~~~
make[4]: *** [Makefile:640: epub-document.lo] Error 1
```
For some reason the build warning will be treat as error in rawhide, but i have to deal with it.
The warning is caused by this commit https://github.com/mate-desktop/atril/commit/39ba86c9d50b614483152335aa22ee83ac2af02e
PR was without CI and build logs !
[...]

PR #600, opened 2024-01-08
epub: Avoid crash when index list has extraneous entry

correctmost:
This commit also fixes an incorrect `sizeof` call detected by AddressSanitizer.

Closes #599

### Testing
- Note: My testing occurred against the 1.26 branch
- I verified the EPUB from #599 can now be opened with Atril
- I verified that AddressSanitizer did not report any memory issues when loading the EPUB

PR #604, opened 2024-01-30
fix a incompatible pointer type warning for gcc14

raveit65:
closes https://github.com/mate-desktop/atril/issues/603

### DIFF
--- a/backend/epub/epub-document.c
+++ b/backend/epub/epub-document.c
@@ -632,7 +632,7 @@ check_mime_type(const gchar* uri,GError** error)
         const gchar* mimetypes[] = {"application/epub+zip", "application/x-booki+zip", NULL};
         guint i;
 
-        for (i = 0; i < g_strv_length (mimetypes); i++) {
+        for (i = 0; i < g_strv_length ((gchar**) mimetypes); i++) {
            if (strcmp(mimeFromFile, mimetypes[i]) == 0)
                 return TRUE;
         }
@@ -643,7 +643,7 @@ check_mime_type(const gchar* uri,GError** error)
             mimeFromFile = ev_file_get_mime_type (uri, TRUE, &err);
             if (mimeFromFile)
             {
-                for (i = 0; i < g_strv_length (mimetypes); i++) {
+                for (i = 0; i < g_strv_length ((gchar**) mimetypes); i++) {
                     if (g_strcmp0(mimeFromFile, mimetypes[i]) == 0)
                         return TRUE;
                 }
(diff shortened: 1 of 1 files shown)


lukefromdc (approved):
Same code as my second earlier test, gave same behavior as master and suppressed the warnings. I actually agree with gcc-14 treating incompatable pointers as an error by default, because my experience is code with that problem very often doesn't run right. This one with just a conflict over const seemed to work OK in master but that's the exception not the rule. Good to fix this instead of relying on compiler behavior.

Applying both this and https://github.com/mate-desktop/atril/pull/600 together stopped ALL the build warnings with default cflags and gcc-13 on Debian Unstable

backend/epub/epub-document.c
```diff
@@ -632,7 +632,7 @@ check_mime_type(const gchar* uri,GError** error)
         const gchar* mimetypes[] = {"application/epub+zip", "application/x-booki+zip", NULL};
         guint i;
 
-        for (i = 0; i < g_strv_length (mimetypes); i++) {
+        for (i = 0; i < g_strv_length ((gchar**) mimetypes); i++) {
```
cwendling, line 635: could also be fixed and optimized using
```suggestion
        for (i = 0; mimetypes[i]; i++) {
```

or
```suggestion
        for (i = 0; i < G_N_ELEMENTS (mimetypes); i++) {
```

But that's kind of beyond the scope of this fix.

cwendling (approved):
LGTM, if that's all there is to fixing the issue :+1:

backend/epub/epub-document.c
```diff
@@ -643,7 +643,7 @@ check_mime_type(const gchar* uri,GError** error)
             mimeFromFile = ev_file_get_mime_type (uri, TRUE, &err);
             if (mimeFromFile)
             {
-                for (i = 0; i < g_strv_length (mimetypes); i++) {
+                for (i = 0; i < g_strv_length ((gchar**) mimetypes); i++) {
```
cwendling, line 646: If you choose this solution, let's apply it here as well
```suggestion
                for (i = 0; mimetypes[i]; i++) {
```

raveit65:
wtf, `commit suggest` from conversation did only add one line.

cwendling:
@raveit65 I didn't add it to the second occurrence because I didn't think it was a necessary change, but merely a possible improvement; and you can only have one suggestion per individual hunk, so I couldn't do both at once.  It's no biggie, but maybe we want to harmonize this in a follow-up :)

raveit65:
I pushed a an update to master :-)

cwendling:
@raveit65 I see :)  Though, I'd rather use the same technique in both instance, rather than 2 different ones

raveit65:
It was a kind of bad website automatism. After i added the first suggestion github marked/closed the conversation as resolved and i thought both suggestions were added.
Than i squashed and merged the PR without checking the commit again.
...good i learnt how a small github detail is working :)

PR #608, opened 2024-04-12
icons: Include higher resolution icons

vkareh:
This only includes higher resolution icons to be loaded automatically by the window manager icon cache, resulting in sharper icons in the alt-tab and wnck views.


lukefromdc (approved):
To my eyes and on a rather small for 4K screen, the difference this time around was very subtle but nothing broke. Tested with marco. We should get a review by someone w a larger (in mm) screen ideally

vkareh:
This also affects when the Alt+Tab icons are a different size. For example:
```
$ gsettings get org.mate.Marco.general icon-size
96
```

lukefromdc:
I make so little use of alt-tab that I don't have a default expectation of what the icons should look like.

PR #609, opened 2024-04-16
Cleanup icons Makefile

cwendling:
Replace full-subdirs with a single Makefile for all icons, making things both simpler and more concise.

Also, properly use Automake variable prefixes instead of doing it manually.

This should not have any functional changes, this is a purely build system change.

---

In addition to simply testing, you can also want to check the installation and distributions are the same.  To do so, I'd suggest something like this *before* and after pulling this branch:
```shell
$ make dist
$ tar tf atril-1.28.0.tar.xz > old-dist.mainfest
$ make install DESTDIR=/tmp/atril
$ find /tmp/atril > old-install.manifest
$ rm /tmp/atril -rf
```
And do the same on this branch (or after pulling it), obviously changing destination files.  Then, diff the old/new manifests and verify the only difference is in the distribution of the removed Makefiles.


lukefromdc (approved):
While I am no expert on autotools makefiles, I did just get a clean test build with this applied. Big simplification and it seems to work fine

PR #614, opened 2024-06-14
Fix .cbr mimetype

### LINKED ISSUE #613: Incorrect mimetype for .cbr comic files
In `configure.ac` the mimetype for .cbr files is wrong: `vnd-comicbook.rar`, should be `vnd.comicbook-rar`. It is correct in `backend/comics/comicsdocument.atril-backend.desktop.in`.

janek37:
Fix #613

See also https://bugs.launchpad.net/ubuntu/+source/atril/+bug/1873998

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -664,7 +664,7 @@ if test "x$enable_tiff" = "xyes"; then
     ATRIL_MIME_TYPES="${ATRIL_MIME_TYPES}image/tiff;"
 fi
 if test "x$enable_comics" = "xyes"; then
-    ATRIL_MIME_TYPES="${ATRIL_MIME_TYPES}application/x-cbr;application/x-cbz;application/x-cb7;application/x-cbt;application/vnd.comicbook+zip;application/vnd-comicbook.rar;"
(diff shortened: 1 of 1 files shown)


lukefromdc (approved):
While I really don't know much about mimetype naming, this builds without issue. A test .cbr file opened fine both with and without this, so not sure what the best test of this is.

janek37:
> While I really don't know much about mimetype naming, this builds without issue. A test .cbr file opened fine both with and without this, so not sure what the best test of this is.

I'm sorry, I should have added more context. The main issue from my perspective is that thumbnails aren't being generated for .cbr files unless I manually edit `/usr/share/thumbnailers/atril.thumbnailer`. The correct mimetype for these files can already be found in `backend/comics/comicsdocument.atril-backend.desktop.in`. Here's the official doc: https://www.iana.org/assignments/media-types/application/vnd.comicbook-rar

lukefromdc:
In my case I don't get a thumbnail with or without the applied, but I am not working with anything near a default install so that may be impossible to diagnose

janek37:
This is a really simple change, and easy to verify (e.g. the IANA link above). Why is it not merged yet?

lukefromdc:
Probably because everyone else on the team has been away

lukefromdc (approved):
I just confirmed this is how the minetype comes up in caja's properties window, will merge

PR #628, opened 2025-02-25
Wayland: stop segfaults on some systems

lukefromdc:
*Put a missed instance of gdk_x11_screen_get_screen_number behind x11-only selectors

### DIFF
--- a/shell/ev-application.c
+++ b/shell/ev-application.c
@@ -419,9 +419,17 @@ on_register_uri_cb (GObject      *source_object,
 	g_variant_builder_add (&builder, "{sv}",
 	                       "display",
 	                       g_variant_new_string (gdk_display_get_name (gdk_screen_get_display (data->screen))));
-	g_variant_builder_add (&builder, "{sv}",
-	                       "screen",
-	                       g_variant_new_int32 (gdk_x11_screen_get_screen_number (data->screen)));
+
+	if (GDK_IS_X11_SCREEN (data->screen)) {
+		g_variant_builder_add (&builder, "{sv}",
+		                     "screen",
+		                      g_variant_new_int32 (gdk_x11_screen_get_screen_number (data->screen)));
+	} else {
+		/*Do not crash on wayland, use the first monitor for now*/
+		g_variant_builder_add (&builder, "{sv}",
+		                     "screen",
+		                      g_variant_new_int32 (0));
+	}
 	if (data->dest) {
 		switch (ev_link_dest_get_dest_type (data->dest)) {
 		case EV_LINK_DEST_TYPE_PAGE_LABEL:


shell/ev-application.c
```diff
@@ -419,9 +419,18 @@ on_register_uri_cb (GObject      *source_object,
 	g_variant_builder_add (&builder, "{sv}",
 	                       "display",
 	                       g_variant_new_string (gdk_display_get_name (gdk_screen_get_display (data->screen))));
-	g_variant_builder_add (&builder, "{sv}",
-	                       "screen",
-	                       g_variant_new_int32 (gdk_x11_screen_get_screen_number (data->screen)));
+
+	if (GDK_IS_X11_DISPLAY (gdk_display_get_default())){
```
cwendling, line 423: Better check the actual screen that will be used:

```suggestion
	if (GDK_IS_X11_SCREEN (data->screen)) {
```

shell/ev-application.c
```diff
@@ -419,9 +419,18 @@ on_register_uri_cb (GObject      *source_object,
...
+		g_variant_builder_add (&builder, "{sv}",
+		                     "screen",
+		                      g_variant_new_int32 (gdk_x11_screen_get_screen_number (data->screen)));
+	}
+	else{
```
cwendling, line 428: style:
```suggestion
	} else {
```

shell/ev-application.c
```diff
@@ -419,9 +419,18 @@ on_register_uri_cb (GObject      *source_object,
...
+		/*Do not crash on wayland, use the first monitor for now*/
```
cwendling, line 429: It's not the monitor but the X screen, which is deprecated in GDK and a X11-only concept so using `1` here -- or even nothing -- is fine I guess.

cwendling (commented):
Actually we can probably drop the whole saving as it's use have already been removed in da4ee916b749bdb1baa1f66c828f884392f8086d

shell/ev-application.c
lukefromdc: Just changed this is a force-push, thanks

shell/ev-application.c
lukefromdc: Implemented this in same force-push

shell/ev-application.c
lukefromdc: Wondering if this should actually be zero: there is only one screen even in x11 since at least 3.22 and it is numbered 0

lukefromdc: Second force-push changes this to 0, tested with no problems on my end

shell/ev-application.c
lukefromdc: Marking resolved as suggestion has been implemented

shell/ev-application.c
lukefromdc: Marking resolved as suggestion has been implemented and the screen number has been set to zero

PR #630, opened 2025-03-30
Clean some deprecations

oz123:
Remove some build warnings, attempt to bring to code to a better state.


lukefromdc (approved):
This builds and runs fine, I specifically tested the dual layout since we make a change there. I see we get rid of the old "stock" icon image code here, as well as what appears to be supporting an earlier change in ev_document_model which which I am not familiar.

Anyway, this works

Issue #631, opened 2025-04-15
atril 1.28.2 segmentation faults when attempting to open an epub book


kendell-clark:
This only affects mate 1.28 and later mate 1.26 isn't affected by this. WHen using atril to read epub files, the following error appears, and then atril segmentation faults with a core dump, which is over 1.4gb and too large to attach here. GDK critical. gdk_window_get_origin assertion 'gdk_is_Window (window) failed

lukefromdc:
The GDK warning may be from after the crash. Can you run atril under gdb with the debug symbols installed, and post the backtrace here?

kendell-clark:
I'll try, I've never run a program under gdb and never posted a backtrace. But I want to help fix this bug

lukefromdc:
Can you post a test file here? I don't have any epubs due to zero use of paid or commercial online media and no accounts with book vendors etc.

kendell-clark:
yes I can. Let me attach the book I'm currently reading

kendell-clark:
it will not allow me to attach an epub file. So I'm going to give instructions on generating an epub file instead. Go to fanfiction.net. Pick any fanfic you want to turn into an epub, and copy that link to the clipboard. Next, head to fichub.net. Paste that link into the edit box and click export. Then click download as epub. That will give you an epub file you can test with atril. I"m sorry I can't just attach an epub file directly but github won't allow me to

PR #632, opened 2025-04-23
epub: Disable thumbnailing sidebar.

mtwebster:
A recent update to libwebkitgtk2gtk-4.1 breaks using WebKitWebView in a GtkOffscreenWindow.

Disable the thumbnail sidebar with epub documents, and disable updating the GtkWindow thumbnail.

ref:
https://github.com/WebKit/WebKit/commit/8d7385fad3ffb31d0c438e6b6a50dffb4390840c

Fixes: #631.

### DIFF
--- a/libview/ev-jobs.c
+++ b/libview/ev-jobs.c
@@ -41,9 +41,6 @@
 #include "ev-debug.h"
 
 #include <gtk/gtk.h>
-#if ENABLE_EPUB
-#include <webkit2/webkit2.h>
-#endif
 #include <errno.h>
 #include <glib/gstdio.h>
 #include <glib/gi18n-lib.h>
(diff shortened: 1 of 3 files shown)


lukefromdc:
I don't use epub documents at all, pretty sure I don't have any as I have no interaction with
commercial or paid electronic media. You might want to post a test file (please not a huge one)
for this

mtwebster:
Sure, these are all public-domain as far as I know...

[test-epubs.zip](https://github.com/user-attachments/files/19893950/test-epubs.zip)

Also, the crash could potentially be obscured by the state of user settings and/or file view metadata (at least for previously-viewed files). To avoid interactions like this, you can run atril (or the file manager) with default settings and without metadata:

```
GSETTINGS_BACKEND=memory GIO_USE_VFS=local atril ...
```
In practice, however, this isn't really necessary here - it will crash either trying to generate a window (xprop) thumbnail or populating the sidebar thumbnail page.

lukefromdc:
Thanks, I was just able to verify that we get this crash in wayland too, about to test the PR code

lukefromdc (approved):
Just tested this and it works. We still get a sidebar with epubs, we just have a text not a thumbnailing sidebar. Thumbnailing sidebar still works on a pdf file

lukefromdc:
Crashes need to get fixed so merged

lukefromdc:
Will create a new release containing this so distros can pick it up

lukefromdc:
We have a new Travis error that blocked a release tarball from being generated. 
`no implicit conversion of nil into String (TypeError)`

PR #639, opened 2025-05-18
CI: Try to use github action

yetist:
Try to use github action


yetist:
Fedora failed about iso-codes.

Disable meson test, meson dist and make check.

raveit65:
Rawhide is development and using it is a risk . Packages can always be missing or broken, or whole the build root is broken before gcc rebuild which happens usually before branching new fedora from rawhide. Before gcc rebuild rawhide is definitely unstable. 
fedora:latest is stable.

...well, your decision.

lukefromdc (approved):
Approving, the test release builds and runs fine locally on Debian. This brings back the ability to release and puts us back in business

oz123 (approved):
yep. travis is dead. github actions are terrible but at least working...

PR #640, opened 2025-05-24
Add configure and meson options to replace mate functions

phileimer:
Allows building `atril` without mate-desktop.
Both configure and meson are modified.
This was only tested on a system without mate-desktop, so it needs reviewing and more testing.
The GTK functions used to replace mate's ones generate compilation warnings because they're deprecated.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -120,7 +120,7 @@ PKG_CHECK_MODULES(BACKEND, cairo >= $CAIRO_REQUIRED gtk+-3.0 >= $GTK_REQUIRED)
 PKG_CHECK_MODULES(FRONTEND_CORE, gtk+-3.0 >= $GTK_REQUIRED gthread-2.0 gio-2.0 >= $GLIB_REQUIRED)
 PKG_CHECK_MODULES(GMODULE, gmodule-2.0 >= $GLIB_REQUIRED)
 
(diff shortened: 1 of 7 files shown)


lukefromdc (commented):
At least in a default build (mate components enabled) we still need the git submodule code for this, no idea if this is being asked for when mate-desktop is not present. Testing builds with mate installed.

The meson build worked, the autotools build failed at configuration with mate-desktop not found.

lukefromdc:
Holding requests for other reviews until the autotools build is fixed

phileimer:
@lukefromdc Could you please try another autotools build with the latest fix?

lukefromdc:
THAT typo will definately do it! retesting with it fixed

lukefromdc:
I see that when MATE is not enabled in these builds, we replace
`mate_image_menu_item_new_with_label()`
with
`gtk_menu_item_new_with_label()`
which does not allow icons to be shown in the menus.

At MATE we have promised not to allow this sort of feature removal by GNOME into MATE, so I don't think we can do this

phileimer:
I agree.
I was just looking for an easy way to have atril working on a non mate system.
But it seems to be too complicated to have a durable solution.
Anyway, thank you for having considered this.

lukefromdc:
Installing the mate-desktop library should not be a problem on a non-MATE system

phileimer:
@cwendling  I just wanted to try atril on my custom system that have no mate component. At the beginning I thought many mate packages may be needed. It turns out I managed to have atril running with a small patch, and without installing anything else. That's why I proposed this PR. But, you're right, if the only package to add on a non mate system is libmate-desktop, it's not worth modifying the build system and code to achieve my initial goal.

PR #641, opened 2025-05-31
Update README.md

Ashraf-wan:
makes it easier to compile by adding apt build-dep to install all necessary dependencies atril needs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,9 @@ Atril Requirements to Run Tests
 
 Atril Compilation
 ==================================================
+* git clone https://github.com/mate-desktop/atril
 * git submodule init
 * git submodule update --remote --recursive
+* apt build-dep -y atril 
 * ./autogen.sh
 * make


luigifab (approved):
it's a nice help (for me too)

lukefromdc:
Note that apt-build is part of apt, which only exists on Debian based distros. We should make a note of that so users of other distros don't wonder why this gets` command not found`

README.md
```diff
@@ -28,7 +28,10 @@ Atril Requirements to Run Tests
 
 Atril Compilation
 ==================================================
+* git clone https://github.com/mate-desktop/atril
 * git submodule init
 * git submodule update --remote --recursive
+* apt build-dep -y atril
+* apt install -y mate-common yelp-tools
```
vkareh: This is Debian-specific. I would just list what the general dependencies are without listing a specific command. Alternatively you can have sections for Debian-based distros, RPM-based distros, etc.

PR #648, opened 2025-08-21
Fixed gir install path variables

### LINKED ISSUE #647: Prefix value is not respected by make install
If Atril is configured with GObject introspection enabled, `make install` places some files under /usr with no regard for prefix value.

**Atril version:**
current master

**Steps to reproduce:**

1. Run autogen.sh with --enable-introspection flag and --prefix set to some path.
2. Run `make` and `make install` to build and install Atril.

**Expected behavior:**
Everything is installed under the path set by --prefix.

**Actual behavior:**
Some files are going to subdirectories in /usr.

E.g. if autogen.sh was run with --enable-introspection and --prefix=/home/andrey/mate/tree, `make install` fails with the following:

```
$ make install

...

Libraries have been installed in:
   /home/andrey/mate/tree/lib

...

 /usr/bin/mkdir -p '/usr/share/gir-1.0'
 /usr/bin/install -c -m 644 AtrilDocument-1.5.0.gir '/usr/share/gir-1.0'
install: cannot create regular file '/usr/share/gir-1.0/AtrilDocument-1.5.0.gir': Permission denied
make[3]: *** [Makefile:1080: install-girDATA] Error 1
make[3]: Leaving directory '/home/andrey/mate/atril/libdocument'
make[2]: *** [Makefile:1242: install-am] Error 2
make[2]: Leaving directory '/home/andrey/mate/atril/libdocument'
make[1]: *** [Makefile:1235: install] Error 2
make[1]: Leaving directory '/home/andrey/mate/atril/libdocument'
make: *** [Makefile:673: install-recursive] Error 1
```

adogadkin:
Changed girdir and typelibsdir to depend upon configured prefix settings, similarly to other MATE projects.
Fixes #647.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -339,38 +339,7 @@ AC_SUBST(PREVIEWER_LIBS)
 # GIR
 # ***
 
-AC_MSG_CHECKING([whether GObject introspection is requested])
-AC_ARG_ENABLE([introspection],
-    AS_HELP_STRING([--enable-introspection],
-                   [Enable GObject introspection]),
-    [enable_introspection=$enableval],
-    [enable_introspection=no])
-AC_MSG_RESULT([$enable_introspection])
-
-G_IR_SCANNER=
-G_IR_COMPILER=
-G_IR_GENERATE=
-GIRDIR=
-GIRTYPELIBDIR=
-
-if test "$enable_introspection" = "yes"; then
-  GOBJECT_INTROSPECTION_REQUIRED=0.6
-  PKG_CHECK_MODULES([GOBJECT_INTROSPECTION],[gobject-introspection-1.0 >= $GOBJECT_INTROSPECTION_REQUIRED])
-
-  G_IR_SCANNER="$($PKG_CONFIG --variable=g_ir_scanner gobject-introspection-1.0)"
-  G_IR_COMPILER="$($PKG_CONFIG --variable=g_ir_compiler gobject-introspection-1.0)"
-  G_IR_GENERATE="$($PKG_CONFIG --variable=g_ir_generate gobject-introspection-1.0)"
-  GIRDIR="$($PKG_CONFIG --variable=girdir gobject-introspection-1.0)"
-  GIRTYPELIBDIR="$($PKG_CONFIG --variable=typelibdir gobject-introspection-1.0)"
-fi
-
-AC_SUBST([G_IR_SCANNER])
-AC_SUBST([G_IR_COMPILER])
-AC_SUBST([G_IR_GENERATE])
-AC_SUBST([GIRDIR])
-AC_SUBST([GIRTYPELIBDIR])
-
-AM_CONDITIONAL([ENABLE_INTROSPECTION],[test "$enable_introspection" = "yes"])
+GOBJECT_INTROSPECTION_CHECK([0.6])
 
 dnl ================== libsynctex ===========================================
 AC_ARG_ENABLE([synctex],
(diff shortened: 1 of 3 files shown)


cwendling (changes requested):
Doesn't something like this work?
```diff
diff --git a/configure.ac b/configure.ac
index 65d1c05d..0e664eb9 100644
--- a/configure.ac
+++ b/configure.ac
@@ -360,8 +360,8 @@ if test "$enable_introspection" = "yes"; then
   G_IR_SCANNER="$($PKG_CONFIG --variable=g_ir_scanner gobject-introspection-1.0)"
   G_IR_COMPILER="$($PKG_CONFIG --variable=g_ir_compiler gobject-introspection-1.0)"
   G_IR_GENERATE="$($PKG_CONFIG --variable=g_ir_generate gobject-introspection-1.0)"
-  GIRDIR="$($PKG_CONFIG --variable=girdir gobject-introspection-1.0)"
-  GIRTYPELIBDIR="$($PKG_CONFIG --variable=typelibdir gobject-introspection-1.0)"
+  GIRDIR='${datadir}'"$($PKG_CONFIG --define-variable=datadir= --variable=girdir gobject-introspection-1.0)"
+  GIRTYPELIBDIR='${libdir}'"$($PKG_CONFIG --define-variable=libdir= --variable=typelibdir gobject-introspection-1.0)"
 fi
 
 AC_SUBST([G_IR_SCANNER])
```

Actually I would say we should use [*introspection.m4*](https://gitlab.gnome.org/GNOME/gobject-introspection/-/blob/main/m4/introspection.m4?ref_type=heads) instead of rolling our own, it hasn't been changed since 2015, so it's probably a safe enough dependency.  The only downside, if it is one, is that we'd depend on it to generate the build system (e.g. a Git build would require *introspection.m4* on the system) -- or we distribute a copy.

At any rate, if this change was to be merged, I'd at least like the definitions for those variables to be removed as well, there's not point in keeping them.

cwendling:
> Yes, it works and is similar to what they do in introspection.m4 you linked. Interestingly though, [in documentation](https://gi.readthedocs.io/en/latest/buildsystems/autotoolsintegration.html) they still suggest defining target path variables as

Oh, right… weird, but that's probably fine then.  Either way you like.

lukefromdc:
> Changed GIR configuration to use [introspection.m4](https://gitlab.gnome.org/GNOME/gobject-introspection/-/blob/main/m4/introspection.m4?ref_type=heads) and [Makefile.introspection](https://gitlab.gnome.org/GNOME/gobject-introspection/-/blob/main/Makefile.introspection?ref_type=heads) provided by gobject-introspection.
> 
> The checks in introspection.m4 enable building introspection support if possible by default, which is different from original implementation always defaulting to --enable-introspection=no. Everything else should stay the same.
> 
> As for shipping these files within atril - after going a bit through other MATE projects I found several of them (e.g. mate-menus, caja ...) using these files as external dependency. All those projects will have to adapt as well, should Gnome remove support for autotools. At that point these files would probably go into mate-common, so for now I think it is better to keep uniformity and let atril expect them to be present as part of gobject-introspection.

Since we already depend on these elsewhere no reason to do differently here. I concur if this is ever a problem they should go once into mate-common instead of being duplicated in many packages

vkareh (commented):
Should we change this to use
```
girdir = $(datadir)/gir-1.0
typelibdir = $(libdir)/girepository-1.0
```
like in the docs (and how other MATE repos seem to do it)?

I don't have a strong opinion either way, but bringing it back up for consistency.

PR #665, opened 2025-11-05
Fix Atril meson build system

cwendling:
* Fix build of the Comics, DVI and Pixbuf backends
* Replace backend [boolean](https://mesonbuild.com/Build-options.html#booleans) options with [features](https://mesonbuild.com/Build-options.html#features): this is technically incompatible, but maintaining compatibility would be highly verbose, and I don't believe the Meson build system is mature enough to require 1:1 compatibility.  Consumers will merely have to change `true` to `enabled` and `false` to `disabled`, which seems easy enough to me.  The main advantage with the change itself is that now backends (but the Pixbuf one) are automatically enabled if their dependencies are met, as we do with the Autotools build system.
* Fix libegg error looking up the desktop file (missing `$prefix` in the path)
* Build and install the Caja extension

### DIFF
--- a/backend/epub/epub-document.c
+++ b/backend/epub/epub-document.c
@@ -1683,7 +1683,7 @@ static void
 epub_document_add_mathJax(gchar* containeruri,gchar* documentdir)
 {
     gchar *containerfilename= g_filename_from_uri(containeruri,NULL,NULL);
-    GString *mathjaxdir = g_string_new("/usr/share/javascript/mathjax");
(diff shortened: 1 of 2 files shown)


cwendling:
@lukefromdc sorry for the delay, but it should now be fixed (plus other improvements related to mathjax)

lukefromdc (approved):
Didn't run internal tests, but the lat change stopped the build failure and I got a successful build that works. I don't know enough abou meson to evaluate the build system code-wise

PR #671, opened 2025-12-23
Fix build with GH actions

oz123:
Building on @yetist work, I fixed some build issue. However, I had to temporarily disable the DVI backend together with meson on latest fedora and ubuntu as there are issues with it. Currently, meson is still in testing so no harm for any users.


lukefromdc:
This fixed the ci issue, thanks. Checked both meson and autotools builds locally and found one unrelated issue: in meson we are not building the help directory for some reason

lukefromdc:
Thanks, I really should have done that myself as soon as I saw this but got lazy..

PR #678, opened 2026-02-13
window: sidebar ignores settings when no metadata

### LINKED ISSUE #163: Side panel hidden by default despite settings.
I have configured atril to show the side panel by default, despite
this it is hidden by default. This behaviour is new.

vkareh:
The sidebar settings were ignored when opening documents on systems without GVFS metadata support.

Now when metadata is not available, we use whatever is in gsettings, and when metadata is available, we respect the per-document state.

Fixes #163


mbkma (approved):
I did not really test, but code looks good and builds fine

PR #681, opened 2026-03-06
Fix zoom limits

### LINKED ISSUE #575: [1.26.0] Max zoom 85, even for 100 or 125% on dropdown, or [1] (100%) icon.
#### Expected behaviour
Be able to zoom to whatever the max zoom is ([400%?](https://github.com/mate-desktop/atril/issues/443)) with ctrl+mousewheel, and/or respect zoom dropdown, and 100% icon, commands.

#### Actual behaviour
Zoom maxes out at 85%. Selecting 100 or 125% on the dropdown doesn't increase the Zoom, but does remain on the selected zoom level, which is incorrect. Having selected 125% on the dropdown, hitting [1] (reset zoom to 100% button) changes zoom back to 85%.

Maybe this is document-specific but zoom past 85% works on a different PDF program.
PDF: https://charts.noaa.gov/PDFs/4149.pdf

#### MATE general version
Xubuntu 22.10

#### Package version
Atril 1.26.0

#### Linux Distribution
Xubuntu 22.10

#### Link to bugreport of your Distribution (requirement)
No posted to distro bugreps, posted here directly.

vkareh:
Currently Atril limits zoom based on the page cache size, configured in gsettings `org.mate.Atril page-cache-size`. Default is 50MB, which prevents large images from zooming in properly. However, increasing that limit too high can crash Atril if you zoom an image to be larger than the surface renderer can handle.

This change backports a few fixes from Evince and adds zoom limit calculations so that the page cache size can be increased safely.

Fixes #575

### DIFF
--- a/data/org.mate.Atril.gschema.xml
+++ b/data/org.mate.Atril.gschema.xml
@@ -26,7 +26,7 @@
       <summary>The URI of the directory last used to save a picture.</summary>
     </key>
     <key name="page-cache-size" type="u">
-      <default>50</default>
+      <default>500</default>
       <summary>Page cache size in MiB</summary>
       <description>The maximum size that will be used to cache rendered pages, limits maximum zoom level.</description>
     </key>
--- a/libview/ev-pixbuf-cache.c
+++ b/libview/ev-pixbuf-cache.c
@@ -430,11 +430,14 @@ ev_pixbuf_cache_get_page_size (EvPixbufCache *pixbuf_cache,
 			       gint           rotation)
 {
 	gint width, height;
+	gint device_scale;
 
+	device_scale = get_device_scale (pixbuf_cache);
 	_get_page_size_for_scale_and_rotation (pixbuf_cache->document,
 					       page_index, scale, rotation,
 					       &width, &height);
-	return height * cairo_format_stride_for_width (CAIRO_FORMAT_RGB24, width);
+	return (height * device_scale) *
+		cairo_format_stride_for_width (CAIRO_FORMAT_RGB24, width * device_scale);
 }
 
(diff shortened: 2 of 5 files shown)


Photon89:
Thanks for the PR, it addresses a very annoying problem!

mbkma:
that is nice, but should we also increase the default value of 50? Because I think most end users will not benefit from this PR, when they need to set the gsettings value themselves.

mbkma (approved):
other than my (non-blocking) question, it is good to go

vkareh:
@mbkma 
> that is nice, but should we also increase the default value of 50? Because I think most end users will not benefit from this PR, when they need to set the gsettings value themselves.

This is a good point. Let's do math! \o/
- A0 paper (the largest standard size I know of) is defined as 1m² ([wikipedia](https://en.wikipedia.org/wiki/International_standard_paper_sizes#A_series))
- A HiDPI monitor uses 2×96 DPI
- Cairo stores pixels as 32 bits (4 bytes): ([cairo docs](https://www.cairographics.org/manual-1.0.2/cairo-Image-Surfaces.html#cairo-format-t))

So for an A0 document on a HiDPI display:
- pixels per meter: 2 × 96 px/in ÷ 0.0254 m/in ≈ 7559 px/m
- pixels per m²: 7559² ≈ 57Mpx
- a single page: 57Mpx × 4bytes/px = 228MiB

Basically means 50 is not even enough to show an A0 page at 100% zoom :facepalm: 

I'm bumping to 500 to be safe...

Photon89:
> I'm bumping to 500 to be safe...

Perrrrfect! :smiley:

PR #683, opened 2026-03-11
fix: do not clear PDF modification flags on save

### LINKED ISSUE #682: Repeated Save As uses unmodified original file and causes data loss
Tested on v1.28.3. Reproduction steps:

- Open `original-form.pdf`. Make changes to the form by filling it in
- Save using "File > Save As". We'll use `original-form-filled.pdf` as the name.
- Once saved, the new file will have the filled form contents.
- Without closing Atril or changing the filled form buffer, repeat "File > Save As" again. Either overwrite `original-form-filled.pdf` or create a new file, doesn't matter.
- The newly created file will have `original-form.pdf`'s contents in it (unfilled), not the filled form from the screen buffer.

At this point, if the user has overwritten `original-form-filled.pdf` in step 4 and closed Atril, the form contents are lost.

Save As should *always* save the buffer on the screen and not reload the original file.

ericswpark:
When using Save As, Atril clears the flags that indicate the PDF has been modified. Unfortunately, on the next save without modifying the document, Atril uses `poppler_document_save_a_copy()`, which does not keep user edits to the form. Thus, the user is led to believe what they are seeing on the screen is saved, only to close and re-open an empty document.

Since we are NOT reloading the original file the flags should not be cleared.

In fact even if we were to change the URI to the newly saved file as indicated in Save As, if the flags are unset then the next Save As invocation will get rid of user edits to the form, so I'm not sure why the flags were ever cleared to begin with.

Fixes #682

### DIFF
--- a/backend/pdf/ev-poppler.cc
+++ b/backend/pdf/ev-poppler.cc
@@ -251,20 +251,14 @@ pdf_document_save (EvDocument  *document,
 	gboolean retval;
 	GError *poppler_error = NULL;
 
-	if (pdf_document->forms_modified || pdf_document->annots_modified) {
-		retval = poppler_document_save (pdf_document->document,
-						uri, &poppler_error);
-		if (retval) {
-			pdf_document->forms_modified = FALSE;
-			pdf_document->annots_modified = FALSE;
-		}
+	retval = poppler_document_save (pdf_document->document,
+					uri, &poppler_error);
+	if (retval) {
+		pdf_document->forms_modified = FALSE;
+		pdf_document->annots_modified = FALSE;
 	} else {
-		retval = poppler_document_save_a_copy (pdf_document->document,
-						       uri, &poppler_error);
-	}
-						       
-	if (! retval)
 		convert_error (poppler_error, error);
+	}
 
 	return retval;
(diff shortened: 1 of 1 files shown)


vkareh (changes requested):
I can confirm the issue and the fix. Thanks!

However, by removing the flags we now have the problem of never being able to close Atril unless you explicitly click on "Close without saving". Even though it's just an annoyance, since data is not lost, having a warning about data loss will be problematic.

The solution here is to remove the conditional entirely, but keep the flags in place, since now Save and Save A Copy both behave similarly.

Evince solved it this way: https://gitlab.gnome.org/GNOME/evince/-/commit/26dc2f52
We should backport that to capture the full behavior fix.

vkareh:
Keep in mind: In their commit, Evince calls `ev_document_set_modified`. We can ignore that since we don't have it and it's used only for cosmetic purposes.

ericswpark:
@vkareh thank you for the review. I've pulled the original patch and adapted it for Atril to preserve authorship information.

vkareh (approved):
This works well and fixes the thing! Thanks :)

PR #684, opened 2026-03-11
fix memleak

mbkma:


shell/ev-window.c
```diff
@@ -2450,6 +2450,7 @@ ev_window_open_uri (EvWindow       *ev_window,
 
 	if (!g_file_is_native (source_file) && !ev_window->priv->local_uri) {
 		ev_window_load_file_remote (ev_window, source_file);
+		g_object_unref (source_file);
```
vkareh: This will unref the file while it's still being copied, since it's an async function. The `ev_window_load_file_remote` call already frees it in the callback inside `g_file_copy_async`.

shell/ev-window.c
```diff
@@ -2654,6 +2655,7 @@ query_remote_uri_mtime_cb (GFile        *remote,
 				   (GAsyncReadyCallback)reload_remote_copy_ready_cb,
 				   ev_window);
 		g_object_unref (target_file);
+		g_object_unref (remote);
```
vkareh: The ready callback `reload_remote_copy_ready_cb` in the `g_file_copy_async` above already frees `remote`, so unreffing it here double-frees it.

vkareh (changes requested):
Two of these changes lead to double-freeing errors.

vkareh (approved):

PR #687, opened 2026-03-30
Add support for digitally signing documents

mbkma:
inspired by https://gitlab.gnome.org/GNOME/evince/-/merge_requests/488

# Testing Digital Signing in Atril

## Prerequisites

Install NSS tools (needed to manage the certificate database that poppler uses):

```bash
sudo apt install libnss3-tools
```

## 1. Create a Self-Signed Test Certificate

Poppler uses the system NSS database at `~/.pki/nssdb`. The database already exists; we just need to add a test certificate.

```bash
# Generate a private key and self-signed certificate with openssl
openssl req -x509 -newkey rsa:2048 -keyout /tmp/test-key.pem \
    -out /tmp/test-cert.pem -days 365 -nodes \
    -subj "/CN=Test Signer/O=Test Org/C=DE"

# Bundle into PKCS#12 format for import
openssl pkcs12 -export -out /tmp/test-signing.p12 \
    -inkey /tmp/test-key.pem -in /tmp/test-cert.pem \
    -name "Test Signer" -passout pass:test123
```

## 2. Import the Certificate into the NSS Database

```bash
# Import the PKCS#12 into the user NSS database
pk12util -i /tmp/test-signing.p12 -d sql:$HOME/.pki/nssdb \
    -W test123 -K ""

# Verify it was imported
certutil -L -d sql:$HOME/.pki/nssdb
```

You should see "Test Signer" in the list.

## 3. Mark the Certificate as Trusted for Signing

```bash
certutil -M -n "Test Signer" -t "u,u,u" -d sql:$HOME/.pki/nssdb
```

## 4. Run Atril and Test

```bash
# Run the freshly built atril
./build/shell/atril ./atril/test/test-links.pdf
```

### Steps in the UI:
1. Open a PDF document
2. Click **File → Digital Signing…**
3. The info bar appears: *"Draw a rectangle to insert a signature field"*
4. Draw a rectangle on the page by clicking and dragging
5. A certificate selection dialog appears — select "Test Signer"
6. Click **Select**
7. A save dialog appears — choose a filename (e.g. `/tmp/signed.pdf`)
8. Click **Save**
9. The signed PDF opens in a new Atril window

### Verify the signature was written:

```bash
pdfsig /tmp/signed.pdf
```
Or use a recent version of Okular which has signing support.


libdocument/ev-document-signatures.c
```diff
@@ -0,0 +1,213 @@
...
+	data->left_font_size = 20.0;
+	data->border_width = 1.5;
+
+	return data;
+}
+
+void
+ev_document_signatures_data_free (EvSignaturesData *data)
+{
+	g_clear_object (&data->certificate_info);
```
vkareh: This should be `g_clear_pointer` like in the next function
```suggestion
	g_clear_pointer (&data->certificate_info, ev_certificate_info_free);
```

backend/pdf/ev-poppler.cc
```diff
@@ -3334,3 +3338,160 @@ pdf_document_document_layers_iface_init (EvDocumentLayersInterface *iface)
 	iface->hide_layer = pdf_document_layers_hide_layer;
 	iface->layer_is_visible = pdf_document_layers_layer_is_visible;
 }
+
+#if POPPLER_CHECK_VERSION(22, 2, 0)
+static PopplerCertificateInfo *
+find_poppler_certificate_info_by_id (EvCertificateInfo *ev_cert_info)
+{
+	GList *certs = poppler_get_available_signing_certificates ();
```
vkareh: This needs to be freed somewhere.

backend/pdf/ev-poppler.cc
```diff
@@ -3334,3 +3338,160 @@ pdf_document_document_layers_iface_init (EvDocumentLayersInterface *iface)
...
+	for (list = pdf_document_get_available_signing_certificates (document); list != NULL; list = list->next) {
+		EvCertificateInfo *cert_info = (EvCertificateInfo *)list->data;
+
+		if (g_strcmp0 (ev_certificate_info_get_id (cert_info), id) == 0) {
+			info = ev_certificate_info_copy (cert_info);
+			break;
+		}
+	}
+
+	g_list_free_full (list, (GDestroyNotify) ev_certificate_info_free);
```
vkareh: Sneaky bug here: you _seem_ to be freeing `list`, except you've been changing the node in the list on every `for` iteration, so `list` no longer refers to the list head. All previous elements in the list are leaked.

backend/pdf/ev-poppler.cc
```diff
@@ -3334,3 +3338,160 @@ pdf_document_document_layers_iface_init (EvDocumentLayersInterface *iface)
...
+
+	poppler_signing_data_set_signature_rectangle (signing_data, signing_rect);
+
+	poppler_document_sign (POPPLER_DOCUMENT (pdf_document->document), signing_data, cancellable, callback, user_data);
+}
+
+static gboolean
+pdf_document_signatures_can_sign (EvDocumentSignatures *document)
+{
+	return TRUE;
```
vkareh: This seems leftover/hacky - you can't sign documents unless they have a certificate...

backend/pdf/ev-poppler.cc
```diff
@@ -3334,3 +3338,160 @@ pdf_document_document_layers_iface_init (EvDocumentLayersInterface *iface)
...
+                              GAsyncReadyCallback   callback,
+                              gpointer              user_data)
+{
+	PdfDocument *pdf_document = PDF_DOCUMENT (document);
+	PopplerColor *color;
+	PopplerSigningData *signing_data = poppler_signing_data_new ();
+	PopplerCertificateInfo *cert_info = find_poppler_certificate_info_by_id (data->certificate_info);
+
+	if (!cert_info)
+		return;
```
vkareh: Should this propagate a `GError`? Without checking that `error == NULL` in `ev_document_signatures_sign`, it will end up calling `on_document_signed` and assume it succeeded...

libview/ev-view.c
```diff
@@ -284,6 +285,8 @@ static void       ev_view_update_primary_selection           (EvView
 /*** Caret navigation ***/
 static void       ev_view_check_cursor_blink                 (EvView             *ev_view);
 
+void ev_view_stop_signature_rect (EvView *view);
```
vkareh: Should this be `static`?

shell/ev-window.c
```diff
@@ -241,6 +242,12 @@ struct _EvWindowPrivate {
 
 	/* Caret navigation */
 	GtkWidget *ask_caret_navigation_check;
+
+	/* Digital signing */
+	GtkWidget         *certificate_listbox;
+	guint              signature_page;
+	EvRectangle       *signature_bounding_box;
+	EvCertificateInfo *signature_certificate_info;
```
vkareh, line 250: These should probably be freed in the dispose function

shell/ev-window.c
```diff
@@ -3323,7 +3344,298 @@ ev_window_cmd_save_as (GtkAction *action, EvWindow *ev_window)
...
+	return ret;
+}
+
+static void
+on_document_signed (GObject      *source_object,
+                    GAsyncResult *result,
+                    gpointer      user_data)
+{
+	char *file = user_data;
+	gchar *uri = g_strdup_printf ("file://%s", file);
```
vkareh: Use `g_filename_to_uri` here

shell/ev-window.c
```diff
@@ -3323,7 +3344,298 @@ ev_window_cmd_save_as (GtkAction *action, EvWindow *ev_window)
...
+	gtk_entry_set_activates_default (GTK_ENTRY (entry), TRUE);
+	gtk_entry_set_visibility (GTK_ENTRY (entry), FALSE);
+	gtk_box_pack_end (GTK_BOX (box), entry, TRUE, TRUE, 6);
+	gtk_widget_show_all (box);
+
+	gtk_dialog_run (GTK_DIALOG (dialog));
+	ret = g_strdup (gtk_entry_get_text (GTK_ENTRY (entry)));
+	gtk_widget_destroy (dialog);
+
+	return ret;
```
vkareh: If user cancels the dialog, this would return an empty string rather than `NULL`

libdocument/ev-document-signatures.h
```diff
@@ -0,0 +1,134 @@
...
+ * General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ */
+
+#if !defined (__EV_ATRIL_DOCUMENT_H_INSIDE__) && !defined (ATRIL_COMPILATION)
+#error "Only <atril-document.h> can be included directly."
+#endif
```
vkareh, line 24: I'm not sure I understand this...

vkareh (changes requested):
I haven't tested this yet, but it seems like a super useful feature for document attestation without having to rely on Adobe.

From this code it doesn't seem like I can yet validate the signatures, only sign, but that could be a logical next phase.

I pointed out in the review a lack of `GError` propagation - this is probably quite important in terms of making sure any issues during signing are surfaced through the UI, especially since it cannot yet validate signatures in Atril, so the user may think they have signed but cannot verify unless they use a different PDF viewer with support for that.

libdocument/ev-document-signatures.h
mbkma: This enforces library users to only include `libdocument/atril-document.h`. It is in all libdocument header files.

L-U-T-i:
I've built the latest atril git snapshot 39f1f91094a301c9c7a15408ac8b0a5d9e5250f7 with this patch applied. I get  File → Digital Signing, but it is greyed out.

I've already had $HOME/.pki/nssdb before (with also my personal certificate, with which I can sign PDF in Adobe Reader). Just to make sure, I've generated and added "Test Signer" certificate as instructed above. "Digital Signing" menu entry is still greyed out.

I've also opened already signed PDF file to try signature verification part, but "Digital Signing" menu entry is greyed out as well.

System is Rocky Linux 9.7.

What might be the reason?

mbkma:
> I've built the latest atril git snapshot [39f1f91](https://github.com/mate-desktop/atril/commit/39f1f91094a301c9c7a15408ac8b0a5d9e5250f7) with this patch applied. I get File → Digital Signing, but it is greyed out.
> 
> I've already had $HOME/.pki/nssdb before (with also my personal certificate, with which I can sign PDF in Adobe Reader). Just to make sure, I've generated and added "Test Signer" certificate as instructed above. "Digital Signing" menu entry is still greyed out.
> 
> I've also opened already signed PDF file to try signature verification part, but "Digital Signing" menu entry is greyed out as well.
> 
> System is Rocky Linux 9.7.
> 
> What might be the reason?

it could be that in Rocky Linux `pobbler` (atrils pdf backend) is build without `nss`. To check you can do `ldd /usr/lib64/libpoppler.so | grep nss` or wherever you pobbler .so file is located.

L-U-T-i:
In RHEL / Fedora, libpoppler.so is a part of devel package (which is normally not needed / installed in user desktops), I guess. I've therfore checked:
```
# locate libpoppler.so
/usr/lib/libpoppler.so.106
/usr/lib/libpoppler.so.106.0.0
/usr/lib64/libpoppler.so.106
/usr/lib64/libpoppler.so.106.0.0
```
and I get:
```
# ldd /usr/lib64/libpoppler.so.106 | grep nss
	libnss3.so => /lib64/libnss3.so (0x00007fe6bc2f4000)
	libnssutil3.so => /lib64/libnssutil3.so (0x00007fe6bc22a000)
```
is this OK?

mbkma:
yes, but maybe you have an older pobbler version? when you setup the build in atril `./autogen.sh` or `meson setup build` it tells you the exact version. In my case:

`run-time dependency poppler-glib found: YES 24.02.0`

You need to have at least version `22.2.0`

L-U-T-i:
I've checked my build log, there is no such line containing "`run-time dependency poppler-glib found: YES <version>`" (there is no line containing `poppler` in `autogen.sh` part), only "`checking for POPPLER... yes`" (in `./configure` part).

But yes, poppler is version 21.01.0 only in Rocky 9.7 at the moment.

Thank you for an explanation / support.

L-U-T-i:
I've backported Rocky 10.1 poppler 24.02.0 to Rocky Linux 9.7, and then atril with this poppler-glib-devel ver. 24.

I've been asked for a nss password, selected a certificate and get "Digital Signing" now. I've also successfully signed a PDF document (at about normal size rectangle I've been warned about it being too small, and made a huge rectangle after, what looks quite silly now...). I don't get any option to verify a signature though (after reopening this signed document, I got a warning that document is signed, but the signature can not be verified...).

L-U-T-i:
I've noticed NSS password is now required whenever I open any PDF. It is quite annoying, as I expect to sign only about 1/1000 of all PDFs opened. Maybe it would be better if NSS password is required only after Digital Signature action (sign or verify) is initiated?

lukefromdc:
This would be unacceptable to those who never sign or interact with signed documents
unless a null passphrase can be set and the interface bypassed.

L-U-T-i:
I'm not sure a null passphrase is an option. Considering I've already had the NSS database in my home directory (before testing that), with my government issued certificate already in it, I guess it is used by some other applications (I'd say Firefox is the first candidate, or some of the dedicated signing applications I have) and there shall be some security to protect such certificates.

Plus, it would be a bit strange to request from atril users to first set a null NSS passphrase before they can normally use atril as by now, or?

lukefromdc:
Then we just have to ensure that this does not ask for a passphrase on encountering an unsigned document,
or leave this out until someone comes up with a fix.

L-U-T-i:
I've been asked for a passphrase also while opening an already signed document...

mbkma:
no worries, I will fix this issue before this PR gets merged :)

mbkma:
Now, the NSS database password prompt appears only when it is actually required:

1. when opening a PDF that contains a signature (for verification), and
2. when signing a document.

If [libsecret](https://gnome.pages.gitlab.gnome.org/libsecret/) is available, the password is cached in the [GNOME Keyring](https://wiki.archlinux.org/title/GNOME/Keyring). This means you only need to enter your password once, the first time the NSS database is unlocked.

L-U-T-i:
It works as expected now, thank you.

It is also possible to draw a small rectangle now. But, if it is too small for the preselected font size, some text is simply cropped (on the right and / or at the bottom). I'm not sure it is OK like that - it would probably be better to reduce the font size (as long as the text can still be anyhow readable) or to refuse signing (request to draw a wider / taller rectangle)?

And, a nice addition, if it would not be too complicated to implement - an old Adobe Reader 11 has the ability to add a drawing (scanned hand written signature) to a digital signature (it is positioned on the left side, and signature text / details right to it). Looks really cool.

Adobe Reader mentioned also adds an Adobe logo as a watermark / background below the signature text, and it looks quite fancy. Atril logo (the one as on Atril icon) watermark would probably make such a signature look really professional... ;-)

mbkma:
> It works as expected now, thank you.
> 
> It is also possible to draw a small rectangle now. But, if it is too small for the preselected font size, some text is simply cropped (on the right and / or at the bottom). I'm not sure it is OK like that - it would probably be better to reduce the font size (as long as the text can still be anyhow readable) or to refuse signing (request to draw a wider / taller rectangle)?
> 
> And, a nice addition, if it would not be too complicated to implement - an old Adobe Reader 11 has the ability to add a drawing (scanned hand written signature) to a digital signature (it is positioned on the left side, and signature text / details right to it). Looks really cool.
> 
> Adobe Reader mentioned also adds an Adobe logo as a watermark / background below the signature text, and it looks quite fancy. Atril logo (the one as on Atril icon) watermark would probably make such a signature look really professional... ;-)

cool :) yeah this are features I want to look into in the future, but to not further complicate this PR, I would add those features in other PRs later on

L-U-T-i:
It seems patch can't be applied after recent git snapshot updates anymore.

It fails with:
```
+ git apply -p1
error: patch failed: libview/ev-view.c:8256
error: libview/ev-view.c: patch does not apply
error: patch failed: libview/ev-view.c:289
error: libview/ev-view.c: patch does not apply
error: patch failed: test/Makefile.am:13
error: test/Makefile.am: patch does not apply

```

PR #688, opened 2026-03-30
fix meson help

mbkma:


vkareh (approved):

lukefromdc:
Just tested this myself, help is not built by default but can be turned on with `-Dhelp_files=true` which mostly works. The help files in C are built and installed but not the translations to other languages. The locales are built however

PR #689, opened 2026-03-31
weblate migration

Jibec:
update pot files
remove transifex configuration
remove "es_ES" as it is the same as "es" and prevent weblate usage
prevent po updates when generating pot files (weblate does that for us)


Jibec:
this PR refers to the migration request done in https://github.com/mate-desktop/mate-terminal/issues/487

PR #692, opened 2026-05-08
Fix and off-by-one error for fit-to-page zoom in dual pane mode

uqs:
In dual pane mode, with a mix of landscape and portrait pages, the wrong zoom level was getting applied to the wrong panes.

For comic book spreads with the right side page being double-wide, this was ok. But when the left-page was double-wide, the extra zoom was applied to the *preceeding* dual pane view (making it extra small) and the actual wider pane got the standard zoom factor applied.

Add debug logging showing the decision being made, which uncoveres the bug.

### DIFF
--- a/libview/ev-view.c
+++ b/libview/ev-view.c
@@ -45,6 +45,7 @@
 #include "ev-view-accessible.h"
 #include "ev-view-private.h"
 #include "ev-view-type-builtins.h"
+#include "ev-debug.h"
 
 enum {
 	SIGNAL_SCROLL,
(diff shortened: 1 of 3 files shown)


libview/ev-view.c
```diff
@@ -7357,6 +7374,13 @@ ev_view_zoom_for_size_dual_page (EvView *view,
 		g_assert_not_reached ();
 	}
 
+	ev_debug_message (DEBUG_ZOOM,
+			  "dual: current_page=%d other_page=%d "
+			  "doc=%.1fx%.1f (after 2x width) "
+			  "target=%dx%d scale=%.4f",
+			  view->current_page + 1, other_page + 1,
```
vkareh, line 7381: You'd need a different case for when `other_page` is `-1` (like `other_page=none` for example), otherwise it will show `other_page=0` which is incorrect and confusing.

vkareh (commented):
This looks correct to me, just a comment on potentially misleading debug output.

libview/ev-view.c
uqs: Hmm, how strongly do you feel about this? IIUC this can only happen on the first/title page of a doc (which is on the right in dual pane mode) and to the left of it there is indeed no page, so "none" would make sense. But I think 0 is also somewhat ok _if_ one understands that this is the 1-indexed page numbers as visible in the sidebar. And yeah, the 0 page then doesn't exist.

vkareh: Yeah that's a valid point. Also, this is just a debug message and wouldn't affect functionality otherwise.

vkareh (approved):

PR #694, opened 2026-05-15
ev-application: Quote user-supplied strings in ev_spawn command line

vkareh:
When spawning a new atril instance for cross-document links, the destination and search parameters from the document were interpolated directly into the command line without shell quoting. Values containing spaces or special characters could be split into separate arguments by the shell parser, potentially being interpreted as unintended flags by the child process.

Apply shell quoting to page label, named destination, and search string values before appending them to the command line, consistent with how other spawn sites in the codebase already handle this.

### DIFF
--- a/shell/ev-application.c
+++ b/shell/ev-application.c
@@ -221,18 +221,22 @@ ev_spawn (const char     *uri,
 	/* Page label or index */
 	if (dest) {
 		switch (ev_link_dest_get_dest_type (dest)) {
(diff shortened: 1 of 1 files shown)


lukefromdc (commented):
Just had a succesful test build and run of this with a PDF file

vkareh:
Thanks @lukefromdc - I've merged the change already and pushed a new release tag for 1.28 and 1.26

PR #700, opened 2026-05-29
epub: use g_strndup for parsing document path

vkareh:
Replace manual byte-by-byte copy loop with g_strndup when extracting the directory component from the epub container's full-path attribute. Also check that the content file exists before attempting to parse it, and handle failures in document title extraction gracefully.


lukefromdc (approved):
The build finished fine, though there were as normal so many deprecation warnings I could not detect any new warnings.

The behavioral change was as stated: the test_longpath.epub file opened from atril in terminal showed a malloc error and atril's gui did not open at all with atril from git master, but atril's gui came up normally and atril was simply unable to open the document with this change applied

lukefromdc:
I just released 1.26.4 with these changes included due to some distros not shipping 1.28
https://github.com/mate-desktop/atril/releases/tag/v1.26.4

PR #702, opened 2026-06-03
Added "Erase Metadata" option

Zzooouhh:
Sometimes I want to open up a document file cleanly from the start, resetting the stored data for last visited page number, zoom level etc. so here is a little option to wipe those metadata clean for the target file before opening it afresh.

### DIFF
--- a/shell/main.c
+++ b/shell/main.c
@@ -44,6 +44,7 @@ static gboolean preview_mode = FALSE;
 static gboolean fullscreen_mode = FALSE;
 static gboolean presentation_mode = FALSE;
 static gboolean unlink_temp_file = FALSE;
+static gboolean reset_metadata = FALSE;
 static gchar   *print_settings;
 static const char **file_arguments = NULL;
 
@@ -69,6 +70,7 @@ static const GOptionEntry goption_options[] =
 	{ "preview", 'w', 0, G_OPTION_ARG_NONE, &preview_mode, N_("Run atril as a previewer"), NULL },
 	{ "find", 'l', 0, G_OPTION_ARG_STRING, &ev_find_string, N_("The word or phrase to find in the document"), N_("STRING")},
 	{ "unlink-tempfile", 'u', G_OPTION_FLAG_HIDDEN, G_OPTION_ARG_NONE, &unlink_temp_file, NULL, NULL },
+	{ "reset-metadata", 'r', 0, G_OPTION_ARG_NONE, &reset_metadata, N_("Resets stored metadata for target file to default values"), NULL },
 	{ "print-settings", 't', G_OPTION_FLAG_HIDDEN, G_OPTION_ARG_FILENAME, &print_settings, NULL, NULL },
 	{ "version", 0, G_OPTION_FLAG_NO_ARG | G_OPTION_FLAG_HIDDEN, G_OPTION_ARG_CALLBACK, option_version_cb, NULL, NULL },
 	{ G_OPTION_REMAINING, 0, 0, G_OPTION_ARG_FILENAME_ARRAY, &file_arguments, NULL, N_("[FILE…]") },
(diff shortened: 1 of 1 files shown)


shell/main.c
```diff
@@ -150,6 +152,47 @@ get_label_from_filename (const gchar *filename)
...
+	clear = g_file_info_new ();
+
+	for (i = 0; attrs[i]; i++) {
+		g_file_info_set_attribute (clear,
+								   attrs[i],
+								   G_FILE_ATTRIBUTE_TYPE_INVALID,
+								   NULL);
+	}
+
+	g_file_set_attributes_async (file,
```
vkareh: This is async, but you don't wait for it to finish before the code continues to opening the file. You should probably use `g_file_set_attributes_from_info` to avoid a race condition.

shell/main.c
```diff
@@ -150,6 +152,47 @@ get_label_from_filename (const gchar *filename)
...
+static void
+erase_metadata_attributes (GFile *file)
+{
+	GFileInfo *info;
+	GFileInfo *clear;
+	gchar **attrs;
+	int i;
+	GError *error = NULL;
+
+	info = g_file_query_info (file, "metadata::*", 0, NULL, &error);
```
vkareh, line 164: This should be `metadata::atril::*` so it doesn't erase metadata from other applications.

shell/main.c
```diff
@@ -150,6 +152,47 @@ get_label_from_filename (const gchar *filename)
...
+	if (!info) {
+		if (error) {
+			g_warning ("%s", error->message);
+			g_error_free (error);
+		}
+		return;
+	}
+
+	attrs = g_file_info_list_attributes (info, "metadata");
```
vkareh, line 173: This function can return `NULL`, which would crash on the `for` loop right below.

shell/main.c
```diff
@@ -69,6 +70,7 @@ static const GOptionEntry goption_options[] =
 	{ "preview", 'w', 0, G_OPTION_ARG_NONE, &preview_mode, N_("Run atril as a previewer"), NULL },
 	{ "find", 'l', 0, G_OPTION_ARG_STRING, &ev_find_string, N_("The word or phrase to find in the document"), N_("STRING")},
 	{ "unlink-tempfile", 'u', G_OPTION_FLAG_HIDDEN, G_OPTION_ARG_NONE, &unlink_temp_file, NULL, NULL },
+	{ "erase-metadata", 'e', 0, G_OPTION_ARG_NONE, &erase_metadata, N_("Resets stored metadata for target file to default values"), NULL },
```
vkareh: I would lean towards `reset-metadata` here, since it's a bit more accurate. You're not just erasing, you're erasing and opening the document, which will set it back to whatever the default is.

vkareh (changes requested):
This is a nice feature, thanks for adding it.

For context to other reviewers, since this is one of the more esoteric features in GLib: GIO stores GVFS metadata on files as you interact with them. You can see any file's metadata with `gio info file.pdf | grep atril` and unset any metadata attribute with `gio set -d file.pdf metadata::atril::<attribute-name>`. The problem is that there's no way to unset them all at once. There's interesting info about this very same issue in Evince here: https://mail.gnome.org/archives/evince-list/2018-July/msg00003.html

shell/main.c
Zzooouhh: This actually was reset-metadata until I had a prompt change of heart a minute before submitting :D Reverted it back to reset-metadata as you suggest.

shell/main.c
Zzooouhh: I couldn't get metadata::atril::* to work here during testing. But now I included a filter for metadata::atril::* in the for (i = 0; attrs[i]; i++) loop and it cleans metadata off the target file properly (as far as I can tell on a quick test).

shell/main.c
Zzooouhh: Good point, changed to g_file_set_attributes_from_info now.

shell/main.c
Zzooouhh: Implemented a null check for attrs now.

shell/main.c
vkareh: Yeah that makes sense - looking at the GFile source code GIO treats anything beyond a top-level namespace as the name of the attribute, so `metadata::` is the namespace and anything beyond that is either a wildcard or the attribute name, so it would see `atril::*` as the name of the attribute rather than a sub-namespace wildcard.

vkareh (approved):
Tested and works very well, thanks!

PR #703, opened 2026-06-03
Added page counter in menubar

Zzooouhh:
I like to keep the Atril UI minimal and hide the toolbar as often as reasonable, but one feature I miss the most from the toolbar is seeing the page count and the current page. This PR includes a small page counter display on the empty right side of the menubar in a "(5 of 132)" type format.

### DIFF
--- a/shell/ev-window.c
+++ b/shell/ev-window.c
@@ -184,6 +184,7 @@ struct _EvWindowPrivate {
 	GtkActionGroup   *bookmarks_action_group;
 	guint             bookmarks_ui_id;
 	GtkUIManager     *ui_manager;
+	GtkAction        *page_counter_action;
 
 	/* Fullscreen mode */
 	GtkWidget *fullscreen_toolbar;
@@ -1161,6 +1162,44 @@ ev_window_sidebar_set_current_page (EvWindow    *window,
 		   ev_sidebar_page_support_document (EV_SIDEBAR_PAGE (bookmarks), document)) {
 		ev_sidebar_set_page (sidebar, bookmarks);
 	}
+}
+
+static void
+update_page_counter (GObject    *object,
+		     GParamSpec *pspec,
+		     gpointer    user_data)
+{
+	EvWindow *ev_window = EV_WINDOW (user_data);
+	gint current_page;
+	gint total_pages;
+	gchar *text;
+
+	/* Safety checks */
+	if (!ev_window)
+		return;
+
+	if (!ev_window->priv->page_counter_action)
+		return;
+
+	if (!ev_window->priv->document)
+		return;
+
+	total_pages = ev_document_get_n_pages(ev_window->priv->document);
+
+	if (!ev_window->priv->model)
+		return;
+
+	current_page = ev_document_model_get_page(ev_window->priv->model) + 1;
+
+	text = g_strdup_printf(_("(%d of %d)"),
+			       current_page,
+			       total_pages);
+
+	G_GNUC_BEGIN_IGNORE_DEPRECATIONS;
+	gtk_action_set_label(ev_window->priv->page_counter_action, text);
+	G_GNUC_END_IGNORE_DEPRECATIONS;
+
+	g_free(text);
 }
 
 static void
@@ -6348,6 +6387,11 @@ ev_window_dispose (GObject *object)
 		priv->action_group = NULL;
 	}
 
+	if (priv->page_counter_action) {
+		g_object_unref (priv->page_counter_action);
+		priv->page_counter_action = NULL;
+	}
+
 	if (priv->view_popup_action_group) {
 		g_object_unref (priv->view_popup_action_group);
 		priv->view_popup_action_group = NULL;
@@ -6391,6 +6435,9 @@ ev_window_dispose (GObject *object)
 	if (priv->model) {
 		g_signal_handlers_disconnect_by_func (priv->model,
 						      ev_window_page_changed_cb,
+						      window);
+		g_signal_handlers_disconnect_by_func (priv->model,
+						      update_page_counter,
 						      window);
 		g_object_unref (priv->model);
 		priv->model = NULL;
@@ -7921,6 +7968,7 @@ ev_window_init (EvWindow *ev_window)
 	GtkAccelGroup *accel_group;
 	GtkCssProvider *css_provider;
 	GError *error = NULL;
+	GtkWidget *counter;
 	GtkWidget *sidebar_widget;
 	GtkWidget *overlay;
 	GObject *mpkeys;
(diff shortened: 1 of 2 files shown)


shell/ev-window.c
```diff
@@ -184,6 +184,7 @@ struct _EvWindowPrivate {
 	GtkActionGroup   *bookmarks_action_group;
 	guint             bookmarks_ui_id;
 	GtkUIManager     *ui_manager;
+	GtkAction        *page_counter_action;
```
vkareh, line 187: You'll need to unref this in the dispose function.

shell/ev-window.c
```diff
@@ -8358,6 +8415,16 @@ ev_window_init (EvWindow *ev_window)
 			  G_CALLBACK (find_bar_visibility_changed_cb),
 			  ev_window);
 
+	/* Update page counter */
+	g_signal_connect (ev_window->priv->model,
+			  "notify::page",
+			  G_CALLBACK (update_page_counter),
+			  ev_window);
+	g_signal_connect (ev_window->priv->model,
```
vkareh, line 8430: Both these signals must be disconnected in the dispose function.

shell/ev-window.c
```diff
@@ -8006,6 +8044,17 @@ ev_window_init (EvWindow *ev_window)
 		gtk_ui_manager_get_accel_group (ev_window->priv->ui_manager);
 	gtk_window_add_accel_group (GTK_WINDOW (ev_window), accel_group);
 
+	ev_window->priv->page_counter_action =
+		g_object_new(GTK_TYPE_ACTION,
+					 "name", "PageCounter",
+					 //~ "label", "(0 of 0)",
```
vkareh: Remove commented code

shell/ev-window.c
```diff
@@ -7761,6 +7762,43 @@ ev_window_media_player_key_pressed (EvWindow    *window,
...
+		return;
+
+	total_pages = ev_document_get_n_pages(ev_window->priv->document);
+
+	if (!ev_window->priv->view)
+		return;
+
+	current_page = ev_document_model_get_page(ev_window->priv->model) + 1;
+
+	text = g_strdup_printf("(%d of %d)",
```
vkareh: This needs to be translated

shell/ev-window.c
```diff
@@ -8058,6 +8107,14 @@ ev_window_init (EvWindow *ev_window)
 			    ev_window->priv->menubar,
 			    FALSE, FALSE, 0);
 
+	GtkWidget *counter;
```
vkareh: This should be hoisted

shell/ev-window.c
```diff
@@ -7761,6 +7762,43 @@ ev_window_media_player_key_pressed (EvWindow    *window,
 	}
 }
 
+static void
+update_page_counter (GObject    *object,
+				    GParamSpec *pspec,
+				    gpointer    user_data)
+{
+	EvWindow *ev_window = EV_WINDOW (user_data);
+	EvDocument *document;
```
vkareh: You're not using this anywhere

shell/ev-window.c
```diff
@@ -7761,6 +7762,43 @@ ev_window_media_player_key_pressed (EvWindow    *window,
...
+
+	if (!ev_window->priv->page_counter_action)
+		return;
+
+	if (!ev_window->priv->document)
+		return;
+
+	total_pages = ev_document_get_n_pages(ev_window->priv->document);
+
+	if (!ev_window->priv->view)
```
vkareh: You're never referencing this, so why check it?

vkareh (changes requested):
This will add a bunch of deprecated warnings. You should probably use `G_GNUC_BEGIN/END_IGNORE_DEPRECATIONS` (like elsewhere in the file).

To be honest, I'm ambivalent about this feature. It seems useful, but there are quirks around it:
- I can see an empty widget block on the top-right when no file is open: Image
- Depending on the text length, the block borders keeps expanding or contracting as I scroll
- The widget is still visible even if the toolbar is visible. I think this might be better as an overlay that shows up *while* the user is scrolling, so that it otherwise doesn't clutter the interface.

Zzooouhh:
The visibility of the empty widget block and its bezels is an interesting point; it seems to be theme dependant - e.g. here's how it looks like with Greybird-dark in XFCE (when no document is open, the widget is completely invisible):
atril2

I hesitated a bit about the redundancy of this page counter with the action widget of the toolbar, but to me this felt like enough of an improvement to try doing anyway. I see two points to consider here: 1. the action widget doesn't merely display the page number, but also allows for navigating it, 2. the action widget is optional and can be stripped off the toolbar by the user even now.

shell/ev-window.c
Zzooouhh: Added g_object_unref (priv->page_counter_action)

shell/ev-window.c
Zzooouhh: Added g_signal_handlers_disconnect_by_func which should disconnect both.

shell/ev-window.c
Zzooouhh: My bad, fixed

shell/ev-window.c
Zzooouhh: Found a translation string used by ev-page-action-widget.c which uses the "(%d of %d)" format, which I've now  included here.

shell/ev-window.c
Zzooouhh: Removed now.

shell/ev-window.c
Zzooouhh: This should've checked for ev_window->priv->model, now corrected.

shell/ev-window.c
Zzooouhh: I've moved the declaration of *counter to the top of the function, is there more that should be done?

vkareh: No that's perfect, thanks!

shell/ev-window.c
```diff
@@ -8058,6 +8116,12 @@ ev_window_init (EvWindow *ev_window)
 			    ev_window->priv->menubar,
 			    FALSE, FALSE, 0);
 
+	counter = gtk_ui_manager_get_widget(ev_window->priv->ui_manager,
+										 "/MainMenu/PageCounter");
```
vkareh: This can probably also benefit from a `G_GNUC_BEGIN/END_IGNORE_DEPRECATIONS` wrapper, but I'm not going to block this effort because of that anyway.

vkareh: Meh, the other calls in the same function don't wrap this function. I'm punting on this.

vkareh:
> I see two points to consider here: 1. the action widget doesn't merely display the page number, but also allows for navigating it, 2. the action widget is optional and can be stripped off the toolbar by the user even now.

This is a valid point. Having an always-on way to show the page number makes sense. I wonder if we want a toggle option in the 'View' menu then? The double page counter when the toolbar is visible can be a bit jarring.

cwendling:
IMHO this is a bit weird. I understand the idea, but I feel like the menubar, although indeed a potentially empty area, is not a logical place for this.  If wanted, maybe in the title bar? *"<document name> (page M of N)"* or such?
And IMO ideally yeah this could be disabled (and would be by default, at least if in the menubar).

PR #704, opened 2026-06-03
Standardize height of action_widget

Zzooouhh:
The action widget is the only one with a nonstandard height, 6px above the others. When it's included, it slightly expands the thickness of the toolbar. This PR mitigates this and keeps the toolbar at a constant height as a result.


vkareh (approved):
It works.

What's strange about it is that the page number entry now looks taller (but the overall toolbar is smaller). Works for me.

PR #705, opened 2026-06-03
Increased recent files display to 10 entries

Zzooouhh:
Recent files now go up to 10 entries... the last entry is labeled with a "0" which doubles as a mnemonic.


vkareh (approved):
Nice and simple. Works well.

PR #706, opened 2026-06-03
Only display supported filetypes in Recent Files list

Zzooouhh:
Presently, Recent Files also displays various files that Atril can't open, e.g. if I try and fail to open a file with an unsupported extension, or when I right click and Save Image As... This PR aims to filter the Recent Files list down to only files supported by Atril.

### DIFF
--- a/libdocument/ev-backends-manager.c
+++ b/libdocument/ev-backends-manager.c
@@ -217,6 +217,20 @@ ev_backends_manager_get_backend_info (const gchar *mime_type)
 	return NULL;
 }
 
+gboolean
+ev_backends_manager_mime_type_supported (const gchar *mime_type)
+{
+	if (ev_backends_manager_get_backend_info (mime_type))
+		return TRUE;
+
+#ifdef ENABLE_PIXBUF
+	if (ev_document_factory_mime_type_supported_by_gdk_pixbuf (mime_type))
+		return TRUE;
+#endif
+
+	return FALSE;
+}
+
 EvDocument *
 ev_backends_manager_get_document (const gchar *mime_type)
 {
--- a/libdocument/ev-backends-manager.h
+++ b/libdocument/ev-backends-manager.h
@@ -39,6 +39,7 @@ typedef struct _EvTypeInfo {
 gboolean    _ev_backends_manager_init                     (void);
 void        _ev_backends_manager_shutdown                 (void);
 
+gboolean    ev_backends_manager_mime_type_supported       (const gchar *mime_type);
 EvDocument  *ev_backends_manager_get_document             (const gchar *mime_type);
 
 EV_DEPRECATED
--- a/shell/ev-window.c
+++ b/shell/ev-window.c
@@ -3055,7 +3055,7 @@ ev_window_setup_recent (EvWindow *ev_window)
                         g_object_unref (icon);
 
 		if (++n_items == 10)
-            break;
+			break;
 	}
 
 	g_list_foreach (items, (GFunc) gtk_recent_info_unref, NULL);
--- a/shell/ev-window.c
+++ b/shell/ev-window.c
@@ -3001,29 +3001,21 @@ ev_window_setup_recent (EvWindow *ev_window)
 		    (gtk_recent_info_is_local (info) && !gtk_recent_info_exists (info)))
 			continue;
 
-		action_name = g_strdup_printf ("RecentFile%u", i++);
-		label = ev_window_get_recent_file_label (
-			(n_items + 1) % 10, gtk_recent_info_get_display_name (info));
-
 		mime_type = gtk_recent_info_get_mime_type (info);
 		if (!mime_type)
 			continue;
 
-		/* Check if Atril has a backend for this MIME type */
-		if (!ev_backends_manager_get_document(mime_type)) {
-		#ifdef ENABLE_PIXBUF
-			/* Allow image backend if pixbuf support enabled */
-			if (!mime_type_supported_by_gdk_pixbuf(mime_type))
-				continue;
-		#else
+		if (!ev_backends_manager_mime_type_supported (mime_type))
 			continue;
-		#endif
-		}
+
+		action_name = g_strdup_printf ("RecentFile%u", i++);
+		label = ev_window_get_recent_file_label (
+			(n_items + 1) % 10, gtk_recent_info_get_display_name (info));
 
 		content_type = g_content_type_from_mime_type (mime_type);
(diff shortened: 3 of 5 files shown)


shell/ev-window.c
```diff
@@ -3005,12 +3005,27 @@ ev_window_setup_recent (EvWindow *ev_window)
...
-                if (content_type != NULL) {
-                        icon = g_content_type_get_icon (content_type);
-                        g_free (content_type);
-                }
+		mime_type = gtk_recent_info_get_mime_type (info);
+		if (!mime_type)
+			continue;
+
+		/* Check if Atril has a backend for this MIME type */
+		if (!ev_backends_manager_get_document(mime_type)) {
```
vkareh: This returns a full document that needs to be unrefed. Also, it's a very heavy function to run 10 times for just a boolean check. Maybe something like `ev_backends_manager_get_backend_info` does what you need without needing the entire thing?

shell/ev-window.c
```diff
@@ -3005,12 +3005,27 @@ ev_window_setup_recent (EvWindow *ev_window)
...
 
-                mime_type = gtk_recent_info_get_mime_type (info);
-                content_type = g_content_type_from_mime_type (mime_type);
-                if (content_type != NULL) {
-                        icon = g_content_type_get_icon (content_type);
-                        g_free (content_type);
-                }
+		mime_type = gtk_recent_info_get_mime_type (info);
+		if (!mime_type)
+			continue;
```
vkareh, line 3006: This `continue` (and the ones below as well) leak some of the variables declared above.

shell/ev-window.c
```diff
@@ -3005,12 +3005,27 @@ ev_window_setup_recent (EvWindow *ev_window)
...
+		#else
+			continue;
+		#endif
+		}
+
+		content_type = g_content_type_from_mime_type (mime_type);
+		if (content_type != NULL) {
+				icon = g_content_type_get_icon (content_type);
+				g_free (content_type);
+		}
```
vkareh, line 3019: Indentation seems off in this block

shell/ev-window.c
```diff
@@ -3005,12 +3005,27 @@ ev_window_setup_recent (EvWindow *ev_window)
...
-                }
+		mime_type = gtk_recent_info_get_mime_type (info);
+		if (!mime_type)
+			continue;
+
+		/* Check if Atril has a backend for this MIME type */
+		if (!ev_backends_manager_get_document(mime_type)) {
+		#ifdef ENABLE_PIXBUF
+			/* Allow image backend if pixbuf support enabled */
+			if (!mime_type_supported_by_gdk_pixbuf(mime_type))
```
vkareh: This function call breaks the build when pixbuf is enabled:
```
../shell/ev-window.c:3016:30: error: implicit declaration of function ‘mime_type_supported_by_gdk_pixbuf’ [-Wimplicit-function-declaration]
 3016 |                         if (!mime_type_supported_by_gdk_pixbuf(mime_type))
      |                              ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

vkareh (changes requested):
This PR is a good idea, but currently has some serious memory leaks and build failures.

shell/ev-window.c
Zzooouhh: Fixed now.

shell/ev-window.c
Zzooouhh: Fixed now.

shell/ev-window.c
Zzooouhh: Good point, although that is a static function. This is now moved into a gboolean checker in ev-backends-manager, which should also solve the pixbuf issue.

shell/ev-window.c
Zzooouhh: Addressed in above point.

shell/ev-window.c
vkareh: Oh nice, good approach.

shell/ev-window.c
```diff
@@ -3048,7 +3055,7 @@ ev_window_setup_recent (EvWindow *ev_window)
                         g_object_unref (icon);
 
 		if (++n_items == 10)
-			break;
+            break;
```
vkareh: There's a diff hunk here leftover from something else, best to remove it

libdocument/ev-backends-manager.c
```diff
@@ -217,6 +217,20 @@ ev_backends_manager_get_backend_info (const gchar *mime_type)
...
 }
 
+gboolean
+ev_backends_manager_mime_type_supported (const gchar *mime_type)
+{
+	if (ev_backends_manager_get_backend_info (mime_type))
+		return TRUE;
+
+#ifdef ENABLE_PIXBUF
+	if (ev_document_factory_mime_type_supported_by_gdk_pixbuf (mime_type))
```
vkareh, line 227: Wait, what is this function? Sounds like you forgot to stage the ev-document-factory files before committing.

Zzooouhh: Right. This should be taken care of now, though I don't know how you'll feel about the ev-document-factory wrapper function.

vkareh: But you're still not including the `ev-document-factory.h` file in here:
```
ev-backends-manager.c: In function 'ev_backends_manager_mime_type_supported':
ev-backends-manager.c:227:13: error: implicit declaration of function 'ev_document_factory_mime_type_supported_by_gdk_pixbuf' [-Wimplicit-function-declaration]
  227 |         if (ev_document_factory_mime_type_supported_by_gdk_pixbuf (mime_type))
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

libdocument/ev-backends-manager.h
```diff
@@ -39,6 +39,7 @@ typedef struct _EvTypeInfo {
 gboolean    _ev_backends_manager_init                     (void);
 void        _ev_backends_manager_shutdown                 (void);
 
+gboolean    ev_backends_manager_mime_type_supported       (const gchar *mime_type);
```
vkareh, line 42: This should also be guarded by `#ifdef ENABLE_PIXBUF`

or better yet, since this whole thing almost ends up being a circular dependency: put `ev_backends_manager_mime_type_supported` in `ev-document-factory.c` instead and then you can get rid of the function wrapper entirely, and there's no need to guard the declaration nor add the `#include`...

PR #709, opened 2026-06-08
Replace hand-made EPUB parsing with libgepub

### LINKED ISSUE #345: use libgepub for epub backend
[libgepub](https://gitlab.gnome.org/gnome/libgepub) may be a good choice for writing ePub backend.

vkareh:
The EPUB backend had a bundled copy of minizip and a lot of hand-written XML parsing code to extract archives to a temp directory, walk the OPF manifest, build the TOC, and serve content to WebKit. This was fragile, hard to maintain, and had a bunch of security issues that seem to be surfacing recently thanks to automated AI scans. This PR is itself assisted by AI.

The PR replaces all of that with [libgepub](https://gitlab.gnome.org/GNOME/libgepub), which handles archive access, metadata, spine navigation, and TOC out of the box. Resources are served directly from the archive to WebKit via an `epub://` URI scheme - no need to extract to a temp directory on disk anymore.

The library doesn't have a lot of projects that use it, but it's actively maintained and included in all the major distros.

Each commit is atomic and tested to ensure there are no regressions or broken builds in between and the code all looks correct and sane to me. Tested with both autotools and meson, both with and without epub backend enabled.

Fixes #345


mbkma:
very nice 👍 I fo not have time to test it right now, but the old epub code always had a bad smell to it

lukefromdc (commented):
I looked over the C files that get changes and are not just deleted with this, and did not see anything inconsistant with replacing handwritten code with an external library. That said, I don't know the library in question or webkit well enough to vouch for the new lines being correct, but I was mostly looking for obvious slop or hallucinated AI output.  

The code change volume is a lot, but much of it is deletions as this replaces a lot of functions previously defined in our own code (and reported to have issues) with the external library.  I don't know the build systems well enough to really evaluate the changes there, but they seem to be simplifications, removing an old dependency, and adding the new one.

On another note, I'm already seeing reports on Mastodon that some folks are making lists of common programs and their last pre-AI versions, and are not updating past those versions. It is thus really important to check over every line on something like this, so as to ensure nothing gets in that we would have rejected if writing the code by hand. In doing so we hopefully can take ownership of and responsablity for the code.

lukefromdc:
The fact that this replaces a lot of code that is vulnerable to adversarial AI analysis renders doing this by some means more important than it would otherwise be. It's an example of the arms race question

vkareh:
> The fact that this replaces a lot of code that is vulnerable to adversarial AI analysis renders doing this by some means more important than it would otherwise be. It's an example of the arms race question

Basically fighting AI with AI :robot:

oz123 (approved):
This is great.

yetist (approved):
Great work. I tested it with an EPUB file, and it works fine.

yetist:
After further testing, I discovered a minor issue: the order of the EPUB chapter list displayed in the sidebar is reversed.

yetist:
> After further testing, I discovered a minor issue: the order of the EPUB chapter list displayed in the sidebar is reversed.

This might be an issue with libgepub, as the returned Table of Contents (TOC) is correct for some files but in reverse order for others.
I tested this patch, and it works; the TOC now displays in the correct order.
```
diff --git a/backend/epub/epub-document.c b/backend/epub/epub-document.c
index 567c4d51..6b66d3ff 100644
--- a/backend/epub/epub-document.c
+++ b/backend/epub/epub-document.c
@@ -288,6 +288,7 @@ build_index_from_gepub_toc (EpubDocument *epub_document)
 {
        GList *toc = gepub_doc_get_toc (epub_document->gepub_doc);
        GList *index = NULL;
+       guint first = 0, last = 0;
 
        for (GList *l = toc; l; l = l->next) {
                GepubNavPoint *np = l->data;
@@ -299,10 +300,17 @@ build_index_from_gepub_toc (EpubDocument *epub_document)
                gint chapter = gepub_doc_resource_uri_to_chapter (
                    epub_document->gepub_doc, np->content);
                node->page = (chapter >= 0) ? chapter : 0;
+               if (first == 0) {
+                   first = node->page;
+               }
+               last = node->page;
 
-               index = g_list_prepend (index, node);
+               index = g_list_append (index, node);
        }
 
+       if (first < last) {
+           return index;
+       }
        return g_list_reverse (index);
 }
 

```

cwendling:
> > After further testing, I discovered a minor issue: the order of the EPUB chapter list displayed in the sidebar is reversed.
> 
> This might be an issue with libgepub, as the returned Table of Contents (TOC) is correct for some files but in reverse order for others. I tested this patch, and it works; the TOC now displays in the correct order.
> 
> ```
> diff --git a/backend/epub/epub-document.c b/backend/epub/epub-document.c
> index 567c4d51..6b66d3ff 100644
> --- a/backend/epub/epub-document.c
> +++ b/backend/epub/epub-document.c
> @@ -288,6 +288,7 @@ build_index_from_gepub_toc (EpubDocument *epub_document)
>  {
>         GList *toc = gepub_doc_get_toc (epub_document->gepub_doc);
>         GList *index = NULL;
> +       guint first = 0, last = 0;
>  
>         for (GList *l = toc; l; l = l->next) {
>                 GepubNavPoint *np = l->data;
> @@ -299,10 +300,17 @@ build_index_from_gepub_toc (EpubDocument *epub_document)
>                 gint chapter = gepub_doc_resource_uri_to_chapter (
>                     epub_document->gepub_doc, np->content);
>                 node->page = (chapter >= 0) ? chapter : 0;
> +               if (first == 0) {
> +                   first = node->page;
> +               }
> +               last = node->page;
>  
> -               index = g_list_prepend (index, node);
> +               index = g_list_append (index, node);
>         }
>  
> +       if (first < last) {
> +           return index;
> +       }
>         return g_list_reverse (index);
>  }
>  
> ```

You should prefer keeping the `g_list_prepend()` because it's a lot cheaper (prepending to a GList is `O(1)` while appending is `O(N)`).  Just don't reverse the list at the end if it's in the inverted order:

```c
if (first < last) /* we built it reversed, so invert it here */
    index = g_list_reverse(index);
return index;
```

Also `page` can legitimately be `0`, so the sorting test should rather be based on whether we've encountered other nodes rather than if it's 0, shouldn't it?
```c
if (index) /* check we're not on the first node */
    last = node->page;
else
    first = node->page;
```

yetist:
> > > After further testing, I discovered a minor issue: the order of the EPUB chapter list displayed in the sidebar is reversed.
> > 
> > 
> > This might be an issue with libgepub, as the returned Table of Contents (TOC) is correct for some files but in reverse order for others. I tested this patch, and it works; the TOC now displays in the correct order.
> > ```
> > diff --git a/backend/epub/epub-document.c b/backend/epub/epub-document.c
> > index 567c4d51..6b66d3ff 100644
> > --- a/backend/epub/epub-document.c
> > +++ b/backend/epub/epub-document.c
> > @@ -288,6 +288,7 @@ build_index_from_gepub_toc (EpubDocument *epub_document)
> >  {
> >         GList *toc = gepub_doc_get_toc (epub_document->gepub_doc);
> >         GList *index = NULL;
> > +       guint first = 0, last = 0;
> >  
> >         for (GList *l = toc; l; l = l->next) {
> >                 GepubNavPoint *np = l->data;
> > @@ -299,10 +300,17 @@ build_index_from_gepub_toc (EpubDocument *epub_document)
> >                 gint chapter = gepub_doc_resource_uri_to_chapter (
> >                     epub_document->gepub_doc, np->content);
> >                 node->page = (chapter >= 0) ? chapter : 0;
> > +               if (first == 0) {
> > +                   first = node->page;
> > +               }
> > +               last = node->page;
> >  
> > -               index = g_list_prepend (index, node);
> > +               index = g_list_append (index, node);
> >         }
> >  
> > +       if (first < last) {
> > +           return index;
> > +       }
> >         return g_list_reverse (index);
> >  }
> >  
> > ```
> 
> You should prefer keeping the `g_list_prepend()` because it's a lot cheaper (prepending to a GList is `O(1)` while appending is `O(N)`). Just don't reverse the list at the end if it's in the inverted order:
> 
> ```c
> if (first < last) /* we built it reversed, so invert it here */
>     index = g_list_reverse(index);
> return index;
> ```
> 
> Also `page` can legitimately be `0`, so the sorting test should rather be based on whether we've encountered other nodes rather than if it's 0, shouldn't it?
> 
> ```c
> if (index) /* check we're not on the first node */
>     last = node->page;
> else
>     first = node->page;
> ```

I am sharing this patch simply to highlight the issue: the order of the TOC returned by `libgepub` can be either ascending or descending, so it needs to be checked and fixed.

This patch may not be the optimal solution; feel free to modify it to suit your specific needs.

vkareh:
Yes, I found a few books in my collection that had the index reversed and this fixed it without affecting the others. Thanks.

I added it to the d6a01f8611cf9bac65b856065906c5c1358f9e7d commit: [diff](https://github.com/mate-desktop/atril/compare/ea6ec0f2f7358427ad839def51417ba21f6df476..b1a648b9583b1368b80b988df3aa0bc0db7dc004)

vkareh:
I fixed the sort order issue directly in libgepub and it got merged already: https://gitlab.gnome.org/GNOME/libgepub/-/merge_requests/21

The sort fix in atril is harmless, though, and guards against using an older version of libgepub.

For me this at least validates that libgepub is actively maintained and receptive to improvements.

lukefromdc:
We've been forced by security issues to add a build dependency to Atril to support Epubs. The question is how distros want to handle this within a minor version.

We could branch 1.28 from before this and version bump master to 1.29 if this is likely to cause distros to stop updating Atril (note that some are still on 1.26) and release a 1.28 version without it. That however opens the door to further security problems cropping up. Alternately distros could build without epub support if they really have to but that's a big feature loss for many.

lukefromdc:
Also note that CI has to be adjusted to account for this, we just got build fails there on arch

PR #711, opened 2026-06-17
Allow inverted colors to be set as default via gsettings

### LINKED ISSUE #710: Have inverted colors by default
### Problem Statement

As the title
Also, amazing work.

### Proposed Solution

some sort of settings cache?

### Alternatives Considered

_No response_

### Use Case

Theme

### Additional Context

_No response_

### Checklist

- [x] I have searched existing issues to avoid duplicates
- [x] I have considered if this feature fits the project scope

vkareh:
Modify inverted colors preference to fall back to GSettings when no per-document metadata exists. This allows users to set inverted colors as the default preference, which will apply to all documents that don't have an explicit per-document preference.

Also added summary and description to all keys in the default schema so they are discoverable in dconf-editor.

Fixes #710

### DIFF
--- a/data/org.mate.Atril.gschema.xml
+++ b/data/org.mate.Atril.gschema.xml
@@ -40,39 +40,63 @@
   <schema id="org.mate.Atril.Default" path="/org/mate/atril/default/" gettext-domain="atril">
     <key name="show-toolbar" type="b">
       <default>true</default>
+      <summary>Show toolbar</summary>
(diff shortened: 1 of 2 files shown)


lukefromdc (approved):
This definately works, though we have to use edit->save current settings as default to save any preferences. A later follow-up PR could add a preferences dialog, which is where people are used to finding app-wide settings

PR #713, opened 2026-06-18
Add allow-links-change-zoom setting

### LINKED ISSUE #712: Fit Width settings does not stay when clicking on Index
### Steps to Reproduce

1. Activate View > Fit Width
2. View > Side pane
3. On the Side pane, select Index
4. Click on any of the sections of the Index, the Fit Width settings goes back to Fit page

This bug does not appear when typing a page number, or when clicking on the Side pane, Thumbnails, nor when navigating the file with the mouse scroll button or keyboard arrows. The Continuous settings does not influence the behaviour (I have it on).

NB: this does not happen on all PDF, maybe only on some which have been processed by pdftk.

### Expected Behavior

Clicking on the Index should keep the Fit Width configuration.

### Actual Behavior

Example of PDF which has the issue.

[L4-VoIP.pdf](https://github.com/user-attachments/files/29095156/L4-VoIP.pdf)

### Error Output

```shell

```

### Version

1.26.2-4

### Distribution

Debian 13 trixie.
Package Version: 1.26.2-4

### Additional Context

_No response_

### Reproducibility

Every time

### Checklist

- [x] I have searched existing issues to avoid duplicates
- [x] I have provided all the requested information above

vkareh:
When enabled links are not allowed to change the current zoom mode and level.

Fixes #712

Backported from https://gitlab.gnome.org/GNOME/evince/-/commit/5d8ef0fa

### DIFF
--- a/data/org.mate.Atril.gschema.xml
+++ b/data/org.mate.Atril.gschema.xml
@@ -34,6 +34,10 @@
       <default>true</default>
       <summary>Show a dialog to confirm that the user wants to activate the caret navigation.</summary>
     </key>
+    <key name="allow-links-change-zoom" type="b">
+      <default>false</default>
+      <summary>Allow links to change the zoom level.</summary>
+    </key>
     <child name="default" schema="org.mate.Atril.Default"/>
   </schema>
 
--- a/libview/ev-view-private.h
+++ b/libview/ev-view-private.h
@@ -174,6 +174,7 @@ struct _EvView {
 	EvPageLayout page_layout;
 	GtkWidget *loading_window;
 	guint loading_timeout;
+	gboolean allow_links_change_zoom;
 
 	/* Common for button press handling */
 	int pressed_button;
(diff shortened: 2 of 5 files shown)


lukefromdc (commented):
I just found that with some PDF's (the one I tested is private and cannot be uploaded here), clicking on a page number in the index still resets the zoom to "fit page" even with this option (which sensibly is the default) turned on

vkareh:
@lukefromdc the option is on to preserve the current behavior (and because the cherry-pick from Evince had that). You need to turn it off to preserve the zoom level.

If it makes more sense I can default it to `false`

lukefromdc:
I would recommend doing so as otherwise this is only accessable via dconf-editor or gsettings from
the command line. I suspect the current behavior is considered a bug but I cannot confirm that, What is
Evince's default?

lukefromdc (approved):
Fixes the problem when set to FALSE (default behavior being links allows to reset it), not sure what the best default is. Possibly expose this in the zoom drop-down menu later?

oz123:
My two cents: make the default false.. This is the least surprising behavior.
Users clicking on links do not expect zoom level to change...

vkareh:
I changed it to `false` as the default. I agree it's better behavior.

